PR #19, opened 2022-12-05
Feature/lazy imports

samwelkanda:
Adds a utility method `lazy_import` that will perform lazy loading of module imports.

### DIFF
--- a/kiota_abstractions/utils.py
+++ b/kiota_abstractions/utils.py
@@ -16,9 +16,9 @@ def lazy_import(name):
     """
     if not name or not isinstance(name, str):
         raise ValueError("Module name must be a valid string")
-        
+
     spec = importlib.util.find_spec(name)
-    
+
     if not spec:
         raise ValueError(f"No spec found for: {name}")
 
--- a/kiota_abstractions/utils.py
+++ b/kiota_abstractions/utils.py
@@ -1,6 +1,7 @@
 import importlib.util
 import sys
 
+
 def lazy_import(name):
 
     spec = importlib.util.find_spec(name)
--- a/kiota_abstractions/utils.py
+++ b/kiota_abstractions/utils.py
@@ -3,7 +3,10 @@ import sys
 
 
 def lazy_import(name):
-    """Lazily imports a python module given its absolute path
+    """Lazily imports a python module given its absolute path.
+    
+    Note: This a utility method for use in Kiota generated code only and not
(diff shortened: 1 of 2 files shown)


kiota_abstractions/utils.py
```diff
@@ -0,0 +1,26 @@
...
+def lazy_import(name):
+    """Lazily imports a python module given its absolute path
+
+    Args:
+        name (str): Absolute path to the module
+
+    Returns:
+        module: The module to be imported
+    """
+    spec = importlib.util.find_spec(name)
```
baywet, line 14: there should probably be some defensive programing in place

kiota_abstractions/utils.py
```diff
@@ -0,0 +1,26 @@
+import importlib.util
+import sys
+
+
+def lazy_import(name):
+    """Lazily imports a python module given its absolute path
```
baywet, line 6: We should probably indicate in the doc comment that this method is only meant for the generated code and that people should not take a dependency on it outside of that scenario.

kiota_abstractions/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+    Args:
+        name (str): Absolute path to the module
+
+    Returns:
+        module: The module to be imported
+    """
+    spec = importlib.util.find_spec(name)
+
+    loader = importlib.util.LazyLoader(spec.loader)
```
baywet, line 16: the case where the loader or module cannot be found should probably be handled as well.

baywet (changes requested):

kiota_abstractions/utils.py
samwelkanda: Thanks. Done in https://github.com/microsoft/kiota-abstractions-python/pull/19/commits/aa4c29726a09d763c66181f694dd13c3bb7be701

kiota_abstractions/utils.py
samwelkanda: Added in https://github.com/microsoft/kiota-abstractions-python/pull/19/commits/aa4c29726a09d763c66181f694dd13c3bb7be701

kiota_abstractions/utils.py
samwelkanda: As long as the spec is found, any further errors with loader and module will be gracefully handled by `importlib`.

baywet (approved):

PR #30, opened 2023-01-16
Task/remove response handler

samwelkanda:
This PR is part of https://github.com/microsoft/kiota/issues/1857

Removes `ResponseHandler` parameter from the  `send` methods in `RequestAdapter` .


kiota_abstractions/_version.py
```diff
@@ -1 +1 @@
-VERSION: str = '0.2.2'
+VERSION: str = '0.2.3'
```
Ndiritu, line 1:
```suggestion
VERSION: str = '0.3.0'
```
Since it's a backwards incompatible change during development?

samwelkanda: Thanks. That makes sense.

Ndiritu (approved):

kiota_abstractions/response_handler.py
```diff
@@ -1,28 +0,0 @@
-from abc import ABC, abstractmethod
-from typing import Callable, Dict, Optional, TypeVar
-
-from .serialization import Parsable, ParsableFactory
-
-NativeResponseType = TypeVar("NativeResponseType")
-ModelType = TypeVar("ModelType")
-
-
-class ResponseHandler(ABC):
```
baywet, line 10: don't you want to keep this interface so you can use it with the new request option introduced in the http package?

baywet (approved):

kiota_abstractions/response_handler.py
samwelkanda: I agree.  Makes sense to keep it to use during instantiation of a response handler option.

Has been retained.

baywet (approved):

PR #35, opened 2023-02-06
Feature/error status code

samwelkanda:
Related to https://github.com/microsoft/kiota/issues/2242

### DIFF
--- a/kiota_abstractions/api_error.py
+++ b/kiota_abstractions/api_error.py
@@ -1,13 +1,16 @@
+from typing import Optional
+
+
 class APIError(Exception):
     """The base class for all API errors."""
 
-    def __init__(self, status_code: int, message: str) -> None:
+    def __init__(self, message: str, status_code: Optional[int] = None) -> None:
         self._response_status_code = status_code
         self._message = message
         super().__init__(self._message)
 
     @property
-    def response_status_code(self) -> int:
+    def response_status_code(self) -> Optional[int]:
         """The HTTP status code of the response that caused the error."""
         return self._response_status_code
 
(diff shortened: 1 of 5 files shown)


kiota_abstractions/api_error.py
```diff
@@ -1,4 +1,16 @@
 class APIError(Exception):
+    """The base class for all API errors."""
 
-    def __init__(self, message: str) -> None:
-        super().__init__(message)
+    def __init__(self, status_code: int, message: str) -> None:
+        self._response_status_code = status_code
+        self._message = message
```
SilasKenneth, line 6: Do we have to explicitly set the message when calling the super constructor?

samwelkanda: Yes, in order to show our custom error message, otherwise the super class will only show a generic error message

Ndiritu (dismissed):

kiota_abstractions/api_error.py
```diff
@@ -1,4 +1,19 @@
+from typing import Optional
+
+
 class APIError(Exception):
+    """The base class for all API errors."""
+
+    def __init__(self, message: str, status_code: Optional[int] = None) -> None:
```
Ndiritu, line 7: I think the `status_code` should be required since every failed API request has a response status code

andrueastman: +1

samwelkanda: Thank. That makes sense. Updated in https://github.com/microsoft/kiota-abstractions-python/pull/35/commits/1f6e9384d9434dbed02cd2f9375d635151086cc6

Ndiritu (approved):

PR #39, opened 2023-02-08
Feature/multi valued headers support

samwelkanda:
Related to https://github.com/microsoft/kiota/pull/2271


kiota_abstractions/request_information.py
```diff
@@ -88,14 +88,29 @@ def url(self, url: Url) -> None:
 
     @property
     def request_headers(self) -> Optional[Dict]:
-        return self.headers
-
-    def add_request_headers(self, headers_to_add: Optional[Dict[str, str]]) -> None:
+        final = {}
+        for key, value in self.headers.items():
+            # Use set to remove duplicates
+            final[key] = ', '.join(list(set(value)))
```
SilasKenneth, line 94:
```suggestion
            final[key] = ', '.join(set(value))
```

This might change to 
```python
 final[key] = ', '.join(value)
```

if we choose to use a set for the map values.

kiota_abstractions/request_information.py
```diff
@@ -45,7 +45,7 @@ def __init__(self) -> None:
         self.query_parameters: Dict[str, QueryParams] = {}
 
         # The Request Headers
-        self.headers: Dict[str, str] = {}
+        self.headers: Dict[str, List[str]] = {}
```
SilasKenneth, line 48: Can't we just use a set instead of a list from this point, so we don't worry about duplicate values?
```suggestion
        self.headers: Dict[str, Set[str]] = {}
```

kiota_abstractions/request_information.py
```diff
@@ -88,14 +88,29 @@ def url(self, url: Url) -> None:
...
+    def add_request_headers(
+        self, headers_to_add: Optional[Dict[str, Union[str, List[str]]]]
+    ) -> None:
         """Adds headers to the request
         """
         if headers_to_add:
-            for key in headers_to_add:
-                self.headers[key.lower()] = headers_to_add[key]
+            for key, value in headers_to_add.items():
+                if key.lower() in self.headers:
```
SilasKenneth, line 104: Could move 
```python
key.lower()
```
to a variable instead of lowering every time.

SilasKenneth (changes requested):

kiota_abstractions/request_information.py
samwelkanda: Updated in https://github.com/microsoft/kiota-abstractions-python/pull/39/commits/4d22efd4a9bdeae2be2067e9deb20e873d4631bb

kiota_abstractions/request_information.py
samwelkanda: See https://github.com/microsoft/kiota-abstractions-python/pull/39/commits/4d22efd4a9bdeae2be2067e9deb20e873d4631bb

kiota_abstractions/request_information.py
samwelkanda: https://github.com/microsoft/kiota-abstractions-python/pull/39/commits/4d22efd4a9bdeae2be2067e9deb20e873d4631bb

SilasKenneth (approved):

andrueastman (approved):

andrueastman (approved):

shemogumbe (approved):

PR #58, opened 2023-04-29
Feature/add response headers

samwelkanda:
Related to https://github.com/microsoft/kiota/issues/2621

### DIFF
--- a/kiota_abstractions/api_error.py
+++ b/kiota_abstractions/api_error.py
@@ -1,4 +1,6 @@
 from typing import Dict
+
+
 class APIError(Exception):
     """The base class for all API errors."""
 
@@ -15,8 +17,8 @@ class APIError(Exception):
 
     @response_headers.setter
     def response_headers(self, value: Dict[str, str]) -> None:
-        self._response_status_code = value
-        
+        self._response_headers = value
+
     @property
     def response_status_code(self) -> int:
         """The HTTP status code of the response that caused the error."""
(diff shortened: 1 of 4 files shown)


kiota_abstractions/api_error.py
```diff
@@ -1,11 +1,24 @@
...
 
-    def __init__(self, message: str, status_code: int) -> None:
+    def __init__(self, message: str, headers: Dict[str, str], status_code: int) -> None:
+        self._response_headers = headers
         self._response_status_code = status_code
         self._message = message
         super().__init__(self._message)
 
+    @property
+    def response_headers(self) -> Dict[str, str]:
```
andrueastman, line 14: Just to confirm, any specific reason the values in the dictionary are not a collection similar to the request headers?
https://github.com/microsoft/kiota-abstractions-python/blob/caf545535dee229c6777b946d59783b0a84ad249/kiota_abstractions/request_information.py#L52

samwelkanda: We use a collection in request information for internal purposes to allow passing multiple values for a key, but eventually we concatenate the value for each key to a single string.

For the response, the native http library will provide the response headers as a dictionary of string keys and values

samwelkanda: Example of response headers
```py
{
    'cache-control': 'private', 
    'transfer-encoding': 'chunked', 
    'content-type': 'application/json; odata.metadata=minimal; odata.streaming=true; IEEE754Compatible=false; 
    charset=utf-8', 
    'content-encoding': 'gzip', 
    'strict-transport-security': 'max-age=31536000', 
   'date': 'Tue, 02 May 2023 11:47:03 GMT'
}
```

andrueastman (approved):

PR #64, opened 2023-05-20
Feature/pydantic models

samwelkanda:
Related to https://github.com/microsoft/kiota/issues/2684

Converts all model classes to Pydantic models by inheriting from `pydantic.BaseModel`


samwelkanda:
Closing in favor of dataclass models, removing the need for external dependency.

PR #67, opened 2023-05-23
Feature/dataclass models

samwelkanda:
Related to https://github.com/microsoft/kiota/issues/2684

Converts model and error classes to python's `dataclasses`.


kiota_abstractions/serialization/parsable.py
```diff
@@ -8,26 +9,31 @@
...
 
-class Parsable(ABC):
+@dataclass
+class Parsable(Protocol):
     """
     Defines a serializable model object.
     """
 
-    @abstractmethod
+    def to_dict(self):
```
andrueastman, line 18: As it looks like `dataclasses` is part of the python runtime and is available, would we really need `to_dict` added to the `Parsable` interface rather than the caller directly calling `asdict(c)`.
As `IParsable` is an interface that has a consistent definition across the languages, I'm afraid that we may get ourselves to propagate the functionality across the languages.

shemogumbe: Actually since we can call `asdict `on the Parsable, would there be a reason to have the `to_dict` mathod, if not specific scenario necessitates it then I think it's fine to not have it

PR #88, opened 2023-06-28
Run the tests in CI

andreaTP:
The tests are passing, but I think it's good to ensure the quality of contributions to always run the tests in CI.


baywet (approved):
As always thanks for the contribution

PR #93, opened 2023-06-30
chore: update the BackingStoreFactorySingleton class

musale:
Ensure that the backing store factory that was set in the first instantiation of the class will be the default during run time.

This changes how the BackingStoreFactory is set [here](https://github.com/microsoft/kiota-http-python/blob/be3f79c61932eddf862519f58ed5d460d2648955/kiota_http/httpx_request_adapter.py#L265-L266):

```python
# from this
if backing_store_factory:
  BackingStoreFactorySingleton.__instance = backing_store_factory # directly accessing an internal, protected member

# to this
if backing_store_factory:
  BackingStoreFactorySingleton(backing_store_factory=backing_store_factory)

# or just this as it will default to InMemoryBackingStoreFactory
BackingStoreFactorySingleton(backing_store_factory=backing_store_factory)
```

I would like some more insight on why this has to be a singleton. This update changes the `BackingStoreFactorySingleton` API and I would love knowing I'm not breaking this upstream in the pipeline.

### DIFF
--- a/tests/store/test_backing_store_factory_singleton.py
+++ b/tests/store/test_backing_store_factory_singleton.py
@@ -19,7 +19,7 @@ class MockTestBackingStore(BackingStore, Generic[T]):
         self.init_complete = True
         self.return_only_changed_values = True
 
-    def get(self, key: str) -> Any | None:
+    def get(self, key: str) -> Any:
         return self.store.get(key, None)
 
     def set(self, key: str, value: Any) -> None:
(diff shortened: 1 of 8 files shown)


tests/store/test_backing_store_factory_singleton.py
```diff
@@ -0,0 +1,83 @@
...
+        self.store.clear()
+
+    def enumerate_(self) -> List[Tuple[str, Any]]:
+        return list(self.store.values())
+
+    def enumerate_keys_for_values_changed_to_null(self) -> List[str]:
+        return list(k for k, v in self.store if v is None)
+
+    def subscribe(
+        self, callback: Callable[[str, Any, Any], None], subscription_id: str | None
```
musale, line 38:
```suggestion
        self, callback: Callable[[str, Any, Any], None], subscription_id: str
```

baywet (approved):

samwelkanda (approved):

PR #100, opened 2023-07-18
feat: add request translator

musale:
Realated to https://github.com/microsoft/kiota-http-python/issues/80

### DIFF
--- a/kiota_abstractions/request_adapter.py
+++ b/kiota_abstractions/request_adapter.py
@@ -137,12 +137,12 @@ class RequestAdapter(ABC, Generic[ResponseType, ModelType, RequestType]):
 
     @abstractmethod
     async def convert_to_native_async(self, request_info: RequestInformation) -> RequestType:
-        """Translates the request information into a native HTTP object.
+        """Translates the request information object into a native HTTP client request object.
 
         Args:
-            request_info (RequestInformation): request object to b converted.
+            request_info (RequestInformation): request information object to be converted.
 
         Returns:
-            RequestType: the type of the native request.
+            RequestType: the natively typed HTTP request of the client.
         """
         pass
--- a/kiota_abstractions/request_adapter.py
+++ b/kiota_abstractions/request_adapter.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from datetime import datetime
 from io import BytesIO
-from typing import Dict, List, Optional, TypeVar
+from typing import Dict, List, Optional, TypeVar, Generic
 
 from .request_information import RequestInformation
 from .serialization import Parsable, ParsableFactory, SerializationWriterFactory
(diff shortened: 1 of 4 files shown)


kiota_abstractions/request_adapter.py
```diff
@@ -1,17 +1,18 @@
...
 from .serialization import Parsable, ParsableFactory, SerializationWriterFactory
 from .store import BackingStoreFactory
 
 ResponseType = TypeVar("ResponseType", str, int, float, bool, datetime, BytesIO)
 ModelType = TypeVar("ModelType", bound=Parsable)
+RequestType = TypeVar("RequestType")
 
 
-class RequestAdapter(ABC):
+class RequestAdapter(ABC, Generic[ResponseType, ModelType, RequestType]):
```
baywet, line 15: as I understand the generic type here would be defined for the whole implementing class. Is there a way in python to define generics that are specific to a method?

baywet: I'm not sure I understand the issue here? Are you saying that the second syntax will raise static analysis issues?
Is there a syntax like ?

```python
# this will raise
def add[T](a, b: int) -> T:
  return a + b
```

musale: The second one will raise static analysis issues (method has to take at least one arg of the generic type). About the syntax, AFAIK, no, that syntax is not supported for functions. For classes though there's something similar where you can do:

```python
T = TypeVar("T")

class A(Generic[T]):
  def add(a, b) -> T:
    return a + b

# then call A with a T type you want to _bind_ to it:

calc = A[int]()
calc.add(1, 2)

calc2 = A[float]()
calc2.add(1.1, 2.2)  # here, a non-float arg will show you an error
```
Throughout the various python sdk codebases I noticed that is not how it's used. However, python resolves the type you want to use so it's not something to worry about. Although I prefer this usage because it's explicit and readable. For example in the `RequestAdapter` for httpx client:

https://github.com/microsoft/kiota-http-python/blob/7f216cda606175033bb6f8dd5d6364530f829855/kiota_http/httpx_request_adapter.py#L30-L33

This would be how it would look since it's subclassing an object with a Generic type(RequestAdapter):

```python
class HttpxRequestAdapter(RequestAdapter[httpx.Request], Generic[ModelType]):
  ...


# OR you can instantiate it like this:
httpx_adapter = HttpxRequestAdapter[httpx.Request]()
...
```
This [PEP 484](https://peps.python.org/pep-0484/) addresses some conventions when it comes to using annotations.

samwelkanda: 100% agree with you Martin. Although I'd argue there's not specific benefit derived from defining generic types for a whole class as opposed to individual methods. Once you bind T to a specific type then you don't need to provide an explicit type during instantiation. So this also works
```py
T = TypeVar("T", bound=int)

class A:
    def add(self, a: T, b: T) -> T:
        return a
    
    def sample(self, x: T, y: T) -> T:
        return self.add(x, y)
```
If the args to `sample` are not of type `T`, then the static type checker will complain.

Also important to note that while these type hints are important for static analysis, they can't be enforced at runtime.

baywet: thanks for the additional context here. Resolving.

samwelkanda (approved):

PR #101, opened 2023-07-19
Chore/backing store support

samwelkanda:
Enables backing store for Python.

Part of https://github.com/microsoft/kiota/issues/2858

- [x] Add a backing store interface in the abstractions
- [x] Add a backed model interface in the abstractions
- [x] Add an in memory backing store implementation in the abstractions
- [x] Add a backing store factory interface in the abstractions
- [x] Implement the factory for the in memory backing store
- [x] Add a singleton for the backing store factory in the abstractions
- [x] Implement factories for proxy
- [x] [implement those factories for the backing store

### DIFF
--- a/kiota_abstractions/serialization/parse_node.py
+++ b/kiota_abstractions/serialization/parse_node.py
@@ -174,8 +174,9 @@ class ParseNode(ABC):
         """
         pass
 
+    @property
     @abstractmethod
-    def get_on_before_assign_field_values(self) -> Callable[[Parsable], None]:
+    def on_before_assign_field_values(self) -> Callable[[Parsable], None]:
         """Gets the callback called before the node is deserialized.
 
         Returns:
@@ -183,31 +184,33 @@ class ParseNode(ABC):
         """
         pass
 
+    @on_before_assign_field_values.setter
     @abstractmethod
-    def get_on_after_assign_field_values(self) -> Optional[Callable[[Parsable], None]]:
-        """Gets the callback called before the node is deserialized.
-
-        Returns:
-            Callable[[Parsable], None]: the callback called before the node is deserialized.
-        """
-        pass
-
-    @abstractmethod
-    def set_on_before_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
+    def on_before_assign_field_values(self) -> Optional[Callable[[Parsable], None]]:
         """Sets the callback called before the node is deserialized.
+
+        Returns:
+            Callable[[Parsable], None]: the callback called before the node is deserialized.
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def on_after_assign_field_values(self) -> Optional[Callable[[Parsable], None]]:
+        """Gets the callback called after the node is deserialized.
+
+        Returns:
+            Callable[[Parsable], None]: the callback called before the node is deserialized.
+        """
+        pass
+
+    @on_after_assign_field_values.setter
+    @abstractmethod
+    def on_after_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
+        """Sets the callback after before the node is deserialized.
 
         Args:
             value (Callable[[Parsable], None]): the callback called before the node is
-            deserialized.
-        """
-        pass
-
-    @abstractmethod
-    def set_on_after_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
-        """Sets the callback called after the node is deserialized.
-
-        Args:
-            value (Callable[[Parsable], None]): the callback called after the node is
             deserialized.
         """
         pass
--- a/kiota_abstractions/serialization/parse_node.py
+++ b/kiota_abstractions/serialization/parse_node.py
@@ -210,7 +210,7 @@ class ParseNode(ABC):
         """Sets the callback called after the node is deserialized.
 
         Args:
-            value (Callable[[Parsable], None]): the callback called before the node is
+            value (Callable[[Parsable], None]): the callback called after the node is
             deserialized.
         """
         pass
--- a/kiota_abstractions/serialization/parse_node.py
+++ b/kiota_abstractions/serialization/parse_node.py
@@ -207,7 +207,7 @@ class ParseNode(ABC):
     @on_after_assign_field_values.setter
     @abstractmethod
     def on_after_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
-        """Sets the callback after before the node is deserialized.
+        """Sets the callback called after the node is deserialized.
 
         Args:
             value (Callable[[Parsable], None]): the callback called before the node is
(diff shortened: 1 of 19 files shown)


tests/store/test_in_memory_backing_store.py
```diff
@@ -0,0 +1,195 @@
...
+    assert len(changed_values) == 1
+    assert  changed_values[0][0] == "colleagues" # Backingstore should detect manager property changed
+    colleagues = changed_values[0][1]
+    assert len(colleagues) == 2
+    assert colleagues[0][0].id == "2f8c9b8c-8ea0-46b2-9c6d-660c9b7bf6af"
+    assert colleagues[0][1].id == "2fe22fe5-1132-42cf-90f9-1dc17e325a74"
+    nested_changed_values = mock_user.colleagues[0].backing_store.enumerate_()
+    assert len(nested_changed_values) == 0
+        
+
```
Ndiritu, line 189: Related to above comment. Perhaps add a test where the mock_user "getter" is invoked & test whether it returns a value only if `return_only_changed_values = True`. (Model serialization relies on getters being called & they should return a value only if it has changed since initialization)

Ndiritu (changes requested):

kiota_abstractions/store/backed_model.py
```diff
@@ -1,17 +1,33 @@
...
     """Defines the contracts for a model that is backed by a store.
     """
+    # Stores model information.
+    backing_store: BackingStore
+
+    def __setattr__(self, prop, val):
+        if not prop == "backing_store":
+            self.__backing_store.set(prop, val)
+            super().__setattr__(prop, self.__backing_store.get(prop))
+        super().__setattr__(prop, val)
```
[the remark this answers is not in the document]
samwelkanda: Actually if you look closely we are overriding the default implementation of `__setattr__` and calling the `set()` method of the backing store instead. Unless the property we are setting is the backing store, in that case we set the property as usual.

tests/store/test_in_memory_backing_store.py
samwelkanda: This is the actual behavior(see previous reply). But I'll add the test.

kiota_abstractions/store/backed_model.py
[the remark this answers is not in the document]
Ndiritu: I think there would be a bug in the scenario of fetching from the API, updating some attributes on an instance and PATCH-ing since during serialization, all model properties will return their initial value which was set during deserialization. The backing store get() needs to run at the point a model property's "get" happens.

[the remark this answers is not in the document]
samwelkanda: > Might we need a __getattr__ override which reads the property value from the backing store instead of "getters" returning the attribute value directly

Added here dfd385f912aadbce8b3587ad162515a5f2baeb40

tests/store/test_in_memory_backing_store.py
samwelkanda: Tests added here dfd385f912aadbce8b3587ad162515a5f2baeb40

kiota_abstractions/serialization/parse_node.py
```diff
@@ -174,40 +174,43 @@ def get_bytes_value(self) -> BytesIO:
...
+            Callable[[Parsable], None]: the callback called before the node is deserialized.
         """
         pass
 
+    @on_after_assign_field_values.setter
     @abstractmethod
-    def set_on_after_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
-        """Sets the callback called after the node is deserialized.
+    def on_after_assign_field_values(self, value: Callable[[Parsable], None]) -> None:
+        """Sets the callback after before the node is deserialized.
```
Ndiritu, line 210:
```suggestion
        """Sets the callback called after the node is deserialized.
```

kiota_abstractions/serialization/parse_node.py
```diff
@@ -174,40 +174,43 @@ def get_bytes_value(self) -> BytesIO:
...
 
         Args:
-            value (Callable[[Parsable], None]): the callback called after the node is
+            value (Callable[[Parsable], None]): the callback called before the node is
```
Ndiritu, line 213:
```suggestion
            value (Callable[[Parsable], None]): the callback called after the node is
```

kiota_abstractions/store/backing_store_parse_node_factory.py
```diff
@@ -21,17 +21,11 @@ def __init__(self, concrete: ParseNodeFactory) -> None:
         """
 
         def func1(x):
-            if isinstance(x, BackedModel):
-                backed_model = x
-                backing_store = backed_model.get_backing_store()
-                if backing_store:
-                    backing_store.set_is_initialization_completed(False)
+            if isinstance(x, BackedModel) and x.backing_store:
```
Ndiritu, line 24: Renaming `func1` and `func2` would help with readability

samwelkanda: Updated in https://github.com/microsoft/kiota-abstractions-python/pull/101/commits/a0101da454f53f7301119c1377efcc5d4d0a1f30

baywet (dismissed):

Ndiritu (approved):

kiota_abstractions/store/backing_store_parse_node_factory.py
```diff
@@ -14,18 +20,12 @@ def __init__(self, concrete: ParseNodeFactory) -> None:
...
-            if isinstance(x, BackedModel):
-                backed_model = x
-                backing_store = backed_model.get_backing_store()
-                if backing_store:
-                    backing_store.set_is_initialization_completed(True)
-
-        super().__init__(concrete, func1, func2)
+        def on_before_deserialization(x):
+            if isinstance(x, BackedModel) and x.backing_store:
+                x.backing_store.is_initialization_completed = False
```
andrueastman, line 25: Out of curiousity, should we be calling `set_is_initialization_completed` here as well? Or are they the same thing?

andrueastman (approved):

baywet (dismissed):

kiota_abstractions/store/backing_store_parse_node_factory.py
samwelkanda: `is_initialization_completed` is set to `True` by default, so this changes it to `False` before deserialization.

andrueastman: Is that the same as calling `set_is_initialization_completed(False)` as we do in `on_after_deserialization`?

SilasKenneth (approved):

kiota_abstractions/store/backing_store_parse_node_factory.py
samwelkanda: Yes

PR #107, opened 2023-08-09
Bugfix/backed model initialization

samwelkanda:
##  SUMMARY

1. Fixes a bug in Python backed models where `is_initialization_completed` property of backing store is set as `True` before default values are set, which results in the values being marked as changed. We set `is_initialization_completed` property to False by default and use the `__post_init__` method of the model classes to set the property to `True`.
2. Updates the `str` method of `APIError` to provide more detailed error messages.

### DIFF
--- a/kiota_abstractions/store/backed_model.py
+++ b/kiota_abstractions/store/backed_model.py
@@ -16,6 +16,9 @@ class BackedModel:
     # Stores model information.
     backing_store: BackingStore
 
+    def __post_init__(self):
+        self.backing_store.is_initialization_completed = True
+
     def __setattr__(self, prop, val):
         if prop == "backing_store":
             super().__setattr__(prop, val)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.7.1] - 2023-08-09
 
 ### Added
-- Added an abstract translator method that should convert a `RequestInformation` object into the native client HTTP request object.
-- Enable backing store for Python.
 
 ### Changed
 - Set the default value for the `is_initialization_completed` parameter in the `InMemoryBackingStore` class to be `False` and use the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.7.1] - 2023-08-09
+
+### Added
+- Added an abstract translator method that should convert a `RequestInformation` object into the native client HTTP request object.
+- Enable backing store for Python.
+
+### Changed
+- Set the default value for the `is_initialization_completed` parameter in the `InMemoryBackingStore` class to be `False` and use the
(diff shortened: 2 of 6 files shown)


kiota_abstractions/store/backed_model.py
```diff
@@ -16,6 +16,9 @@ class BackedModel:
     # Stores model information.
     backing_store: BackingStore
 
+    def __post_init__(self):
```
baywet, line 19: this is usually done by the poxy.
Also, I don't know much about python, but this should be an interface (unless python allows for multi parents inheritance)

samwelkanda: > Also, I don't know much about python, but this should be an interface (unless python allows for multi parents inheritance)
It does e.g `class Entity(AdditionalDataHolder, BackedModel, Parsable):`

> this is usually done by the poxy.
Yes, after serialization. The issue here is during instantiation. Instead of `is_initialization_completed` being `True` by default in the backing store, we want it set to T`rue` only after initialization is completed using the `__post_init__` method. That way we can track changes to the fields thereafter.

samwelkanda: The same applies to Python: https://github.com/microsoft/kiota-abstractions-python/blob/554697a71fdaa92e9479808877dc244c0b89ed60/kiota_abstractions/store/backing_store_parse_node_factory.py#L23

The bug is unique to Python(because of how it instantiates dataclasses.) and occurs at https://github.com/microsoft/kiota-abstractions-python/blob/554697a71fdaa92e9479808877dc244c0b89ed60/kiota_abstractions/store/in_memory_backing_store.py#L23 

Say you have a class `Event` which is a backed model

```py
@dataclass
class Event(BackedModel):
    additional_data: Dict[str, Any] = field(default_factory=dict)
    id: Optional[str] = None
    odata_type: Optional[str] = "#microsoft.graph.event"
    attachments: Optional[List[Attachment]] = None
    attendees: Optional[List[Attendee]] = None
    body: Optional[ItemBody] = None
    calendar: Optional[Calendar] = None
    end: Optional[DateTimeTimeZone] = None
    has_attachments: Optional[bool] = None
    subject: Optional[str] = None
```
Currently, during instantiation, Python will first set the backing store property in the base class `BackedModel` and then proceed to set the default values in the `Event` class in order. What this means is that all these properties will be marked as changed since the backing store was already set and `is_initialization_completed` set to True by default.
Thus:

```cmd
event = Event()
event.backing_store.return_only_changed_values = True
event.backing_store.enumerate_()
```
Returns: `[('additional_data', {}),('id', None), ('odata_type', "#microsoft.graph.event"),('attachments', None), ('attendees', None), ('body', None), ('calendar', None), ('end', None), ('has_attachments', None), ('subject', None)]`

This is not correct since no value has changed after instantiation. 

What am proposing is, in the backing we set the default value of `is_initialization_completed` to False. The we wait until initialization has completed and the default values have been set then we use the `__post_init__` method to set it to True. That way 

```cmd
event = Event()
event.backing_store.return_only_changed_values = True
event.backing_store.enumerate_()
```
Returns: `[]` since no value has changed

And 

``cmd
event = Event()
event.subject = "Hello World"
event.backing_store.return_only_changed_values = True
event.backing_store.enumerate_()
```
Returns: `[('subject', "Hello World")]` since subject has changed after initialization.

baywet: alright, thanks for the additional context, I'm guessing this is due to the way Python instantiates objects that varies from other languages.

baywet (approved):

PR #155, opened 2023-10-04
Add a try_add method for request headers

andreaTP:
Preparation work for: https://github.com/microsoft/kiota/pull/3401


SilasKenneth (dismissed):

SilasKenneth (approved):

baywet:
Thanks for the work here @samwelkanda for release
Also this made me realize the Python implementation is not aligned with other languages so I created this issue.
https://github.com/microsoft/kiota/issues/3430

andreaTP:
If we can get this one merged I can proceed with the final implementation in Kiota without risk of breaking 🙏

PR #158, opened 2023-10-11
- adds content type parameter to set stream content

baywet:
related https://github.com/microsoft/kiota/issues/3377

### DIFF
--- a/kiota_abstractions/request_information.py
+++ b/kiota_abstractions/request_information.py
@@ -227,13 +227,17 @@ class RequestInformation:
                 writer_func(None, values)
             self._set_content_and_content_type_header(writer, content_type)
 
-    def set_stream_content(self, value: BytesIO) -> None:
+    def set_stream_content(self, value: BytesIO, content_type: Optional[str]) -> None:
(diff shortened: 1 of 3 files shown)


kiota_abstractions/request_information.py
```diff
@@ -227,13 +227,17 @@ def set_content_from_scalar(
                 writer_func(None, values)
             self._set_content_and_content_type_header(writer, content_type)
 
-    def set_stream_content(self, value: BytesIO) -> None:
+    def set_stream_content(self, value: BytesIO, content_type: Optional[str]) -> None:
```
Ndiritu, line 230: Might need to default to binary or minor rev

baywet: so that's where my lack of experience with python shows.
I'm defaulting the value below. And I thought Optional[str] meant that one could call the method without passing the content type argument?

baywet:
```suggestion
    def set_stream_content(self, value: BytesIO, content_type: str = None) -> None:
```

PR #162, opened 2023-10-17
Chore/refactor request headers

samwelkanda:
Adds a dedicated  `RequestHeaders` class to handle headers

Related to https://github.com/microsoft/kiota/issues/3430

closes #158

### DIFF
--- a/kiota_abstractions/request_headers.py
+++ b/kiota_abstractions/request_headers.py
@@ -143,9 +143,10 @@ class RequestHeaders():
         values = self.try_get(header_name)
         if values:
             values.remove(header_value)  #type: ignore
-            if not values:
-                return self.remove(header_name)
-            return values
+            if bool(values):
+                return values
+            return self.remove(header_name)
+            
         return False
 
     def remove(self, header_name: str) -> Union[bool, Set[str]]:
--- a/kiota_abstractions/request_headers.py
+++ b/kiota_abstractions/request_headers.py
@@ -1,0 +1,184 @@
+from __future__ import annotations
+
+from typing import Dict, List, Set, Union
+
+
+class RequestHeaders():
+    "Represents a collection of request headers"
+
+    def __init__(self) -> None:
+        self._headers: Dict[str, Set[str]] = {}
+        self._single_value_headers: Set[str] = {
+            "content-type", "content-encoding", "content-length"
+        }
+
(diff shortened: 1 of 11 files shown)


kiota_abstractions/request_information.py
```diff
@@ -95,46 +96,25 @@ def url(self, url: Url) -> None:
...
-    def try_add_request_header(self, key: str, value: str) -> bool:
-        """Try to add an header to the request if it's not already set"""
-        if key and value:
-            lowercase_key = key.lower()
-            if lowercase_key in self.headers:
-                return False
-            self.headers[lowercase_key] = {str(value)}
-            return True
-        return False
+    def add_request_headers(self, headers_to_add: RequestHeaders) -> None:
```
baywet, line 103: is there much value in keeping those two vanity methods? Python would be the only language with those.

kiota_abstractions/request_headers.py
```diff
@@ -0,0 +1,185 @@
...
+
+from typing import Dict, List, Set, Union
+
+
+class RequestHeaders():
+    "Represents a collection of request headers"
+
+    def __init__(self) -> None:
+        self._headers: Dict[str, Set[str]] = {}
+        self._single_value_headers: Set[str] = {
```
baywet, line 11: could this be static instead? we don't need an instance of those per headers class instance.

kiota_abstractions/request_headers.py
```diff
@@ -0,0 +1,185 @@
+from __future__ import annotations
+
+from typing import Dict, List, Set, Union
+
+
+class RequestHeaders():
```
baywet, line 6: Maybe call that class headers collection instead since it can be used for response headers?

samwelkanda: Makes sense. Updated.

kiota_abstractions/request_information.py
samwelkanda: Not really needed. Removed.

kiota_abstractions/request_headers.py
samwelkanda: I'm not sure I follow. I assumed we create `HeadersCollection` instances and track changes per instance. No?

baywet: I was referring to the collection of single valued headers.

baywet (approved):
Thanks for making the changes!

PR #219, opened 2024-01-31
Feature/generator reduction

samwelkanda:
## Overview

Part of https://github.com/microsoft/kiota/issues/3695


kiota_abstractions/request_configuration.py
```diff
@@ -4,14 +4,15 @@
...
+from typing import Dict, List, Optional
 
+from .default_query_parameters import QueryParameters
 from .headers_collection import HeadersCollection
 from .request_option import RequestOption
 
 
 @dataclass
-class BaseRequestConfiguration:
+class RequestConfiguration:
```
baywet, line 15: same concern about this rename here

tests/conftest.py
```diff
@@ -117,7 +117,7 @@ def mock_user():
 
 @pytest.fixture
 def mock_request_information():
-    request_info = RequestInformation()
+    request_info = RequestInformation(None, None)
```
baywet, line 120: could we make those parameters optional somehow so people can use request information without having to configure the request information right away should they need to?
Plus adding required parameters is a source breaking change

baywet (changes requested):

kiota_abstractions/base_request_configuration.py
```diff
@@ -20,3 +21,5 @@ class BaseRequestConfiguration:
 
     # Request options
     options: Optional[List[RequestOption]] = None
+
+    query_parameters: Optional[GetQueryParameters] = None
```
baywet, line 25: My suggestion here if that makes sense in python:
- Add a QueryParameters class that has everything GetQueryParameters had
- Mark GetQueryParameters as deprecated/obsolete
- Empty GetQueryParameters
- Make GetQueryParameters inherit from QueryParameters
- Update all references to GetQueryParameters to QueryParameters in the libs but also in the generation

This way older clients that pull the updated dependency will still work (but get a warning hopefully)
Newer clients get the correct naming.
We can delete GetQueryParameters in v2.


Thoughts?

samwelkanda: Yeah, this should work. Will be both backward compatible and make python aligned with other languages in terms of naming.

kiota_abstractions/base_request_configuration.py
```diff
@@ -4,8 +4,9 @@
 # See License in the project root for license information.
 # ------------------------------------
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional
 
+from .default_query_parameters import QueryParameters
 from .headers_collection import HeadersCollection
 from .request_option import RequestOption
```
baywet, line 11: as far as I understand, we're not going to be generating request configuration classes for python anymore, correct?

In that case, shouldn't the query parameters be generic so people get proper auto completion?

if we do stop generating request configuration classes, I suggest the same derive, name, deprecate trick as query parameters to clean things up.

samwelkanda: > as far as I understand, we're not going to be generating request configuration classes for python anymore, correct?

That was the plan, in line with languages like dotnet. Is this a requirement or can we keep them for python?

> In that case, shouldn't the query parameters be generic so people get proper auto completion?

Currently it's just an empty class with no specific fields/parameters. Or is the issue naming?

> if we do stop generating request configuration classes, I suggest the same derive, name, deprecate trick as query parameters to clean things up.

Why deprecate it in abstractions? My understanding is that we deprecate the generated ones in favor of using the generic `RequestConfiguration` class in abstractions

baywet: > That was the plan, in line with languages like dotnet. Is this a requirement or can we keep them for python?

The goal here is to reduce the generated code as much as possible without impacting the developer experience or the source compatibility. Anything we can avoid generating within those requirements, we should.

> Currently it's just an empty class with no specific fields/parameters. Or is the issue naming?

My concern here is that since we're not using generic parameters when they type the Python equivalent of
`client.users.get(o => o.QueryParameters.Select)`, they won't get auto-completion of Select here.

> Why deprecate it in abstractions?

My guess is that the currently generated code (without your recent changes) depends on BaseRequeestConfiguration. So removing it or changing it too much would be source breaking. But to accommodate for the generic query parameters type value, I'm suggesting to introduce a new class, and deprecate this one.

Hopefully all of this makes sense now?

samwelkanda: Gotcha. Deprecated `BaseRequestConfiguration` in favor of  the new `RequestConfiguration` class. Also added  `$filter`, `$select`, `$orderby`, `$count`, `$top`, `$skip` and $`expand` as generic parameters

kiota_abstractions/default_query_parameters.py
```diff
@@ -12,7 +13,26 @@ class QueryParameters:
...
+    orderby: Optional[List[str]] = None
+
+    # Select properties to be returned
+    select: Optional[List[str]] = None
+
+    # Skip the first n items
+    skip: Optional[int] = None
+
+    # Show only the first n items
+    top: Optional[int] = None
```
baywet, line 35: I think we should remove those as :
- this makes an assumption of an OData API
- this makes the assumption those query parameters are supported which could lead the application developer to think so when it's not the case

kiota_abstractions/base_request_configuration.py
```diff
@@ -23,3 +24,11 @@ class BaseRequestConfiguration:
     options: Optional[List[RequestOption]] = None
 
     query_parameters: Optional[QueryParameters] = None
+
+
+@dataclass
+class BaseRequestConfiguration:
```
baywet, line 30: I believe it should inherit from RequestConfiguration to maintain compatibility with older generated clients

baywet (changes requested):

kiota_abstractions/default_query_parameters.py
samwelkanda: I concur on the two points.

> My concern here is that since we're not using generic parameters when they type the Python equivalent of
client.users.get(o => o.QueryParameters.Select), they won't get auto-completion of Select here.

But now I'm not sure what you mean by `generic parameters`.

baywet: By that I mean a type parameter for the request option class, or a template if we talk C++.

baywet (approved):

PR #235, opened 2024-02-28
Feature/multipart serialization

samwelkanda:
## Overview

Adds multipart body class to support serialization of multipart request bodies.

## Related Issue
https://github.com/microsoft/kiota/issues/3030

### DIFF
--- a/kiota_abstractions/request_information.py
+++ b/kiota_abstractions/request_information.py
@@ -77,7 +77,7 @@ class RequestInformation:
         self.headers: HeadersCollection = HeadersCollection()
 
         # The Request Body
-        self.content: Optional[BytesIO] = None
+        self.content: Optional[bytes] = None
 
     def configure(self, request_configuration: RequestConfiguration) -> None:
         """Configures the current request information headers, query parameters, and options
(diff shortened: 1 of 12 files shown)


kiota_abstractions/request_information.py
```diff
@@ -145,8 +153,8 @@ def remove_request_options(self, options: List[RequestOption]) -> None:
 
     def set_content_from_parsable(
         self,
-        request_adapter: Optional["RequestAdapter"],
-        content_type: Optional[str],
+        request_adapter: RequestAdapter,
```
baywet, line 156: wouldn't these changes be source breaking?

kiota_abstractions/serialization/serialization_writer.py
```diff
@@ -146,7 +146,7 @@ def write_collection_of_enum_values(
         pass
 
     @abstractmethod
-    def write_bytes_value(self, key: Optional[str], value: BytesIO) -> None:
+    def write_bytes_value(self, key: Optional[str], value: bytes) -> None:
```
baywet, line 149: wouldn't these changes be source breaking?

kiota_abstractions/request_information.py
samwelkanda: Not really a breaking change from an end user perspective as the type annotations in Python are not enforced at runtime, they are merely hints. However, we'll need to update the signature in the concrete classes for consistency.

baywet (approved):

baywet (approved):

PR #278, opened 2024-05-21
Adds path/query parameter sanization

andrueastman:
Fixes https://github.com/microsoft/kiota-abstractions-python/issues/245

Sanitizes types passed to the path/query object dictionaries to ensure success of various object types in the parameters. 
Adds tests to validate date,time,datetime,boolean,number and uuid types.

### DIFF
--- a/kiota_abstractions/request_information.py
+++ b/kiota_abstractions/request_information.py
@@ -303,6 +303,12 @@ class RequestInformation:
             return ','.join([x.value for x in value])
         if isinstance(value, UUID):
             return str(value)
+        if isinstance(value, datetime):
+            return value
+        if isinstance(value, date):
+            return str(value)
+        if isinstance(value, time):
+            return str(value)
         return value
 
     def _decode_uri_string(self, uri: Optional[str]) -> str:
--- a/kiota_abstractions/request_information.py
+++ b/kiota_abstractions/request_information.py
@@ -109,10 +109,10 @@ class RequestInformation:
 
         data: Dict[str, Any] = {}
         for key, val in self.query_parameters.items():
-            val = self._replace_enum_values_with_string_representation(val)
+            val = self._get_sanitized_value(val)
             data[key] = val
         for key, val in self.path_parameters.items():
-            val = self._replace_enum_values_with_string_representation(val)
+            val = self._get_sanitized_value(val)
             data[key] = val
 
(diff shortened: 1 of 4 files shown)


baywet (commented):
I think the version in the toml file at the root also needs to be updated

baywet:
interesting, it's not configured like [the service lib](https://github.com/microsoftgraph/msgraph-sdk-python/blob/6abd71affd2f37da21e89c3aa605fc5718eb9000/pyproject.toml#L7)

I believe it's because of a combination of [this file](https://github.com/microsoft/kiota-abstractions-python/blob/3c294832dfbf87a75edbe3e5ddb24b7a240a9050/kiota_abstractions/__init__.py#L4) and [that directive](https://github.com/microsoft/kiota-abstractions-python/blob/3c294832dfbf87a75edbe3e5ddb24b7a240a9050/pyproject.toml#L24) but maybe @samwelkanda can confirm?
If so, it'd be interesting to replicate in other python repos, for consistency and efficiency.

kiota_abstractions/request_information.py
```diff
@@ -291,17 +291,26 @@ def _set_content_and_content_type_header(
...
+        elif isinstance(value, list) and all(isinstance(x, Enum) for x in value):
+            sanitized_value = ','.join([x.value for x in value])
+        elif isinstance(value, UUID):
+            sanitized_value = str(value)
+        elif isinstance(value, datetime):
+            sanitized_value = value
+        elif isinstance(value, date):
+            sanitized_value = str(value)
+        elif isinstance(value, time):
+            sanitized_value = str(value)
```
musale, line 312:
```suggestion
        elif any([isinstance(value, UUID), isinstance(value, date), isinstance(value, time)]:)
            sanitized_value = str(value)
        elif isinstance(value, datetime):
            sanitized_value = value
```
Collapse the `str`ingers.

samwelkanda:
@baywet @andrueastman We do not need to update the `pyproject.toml` file for now as the core libraries use `flit` to publish packages to PyPI which only requires the VERSION variable to be updated. I agree that we should update the workflows to align with the setup in service libraries that use Python’s built-in `build` tool. @shemogumbe are you able to take this up?

pjmagee:
Aweomse stuff

shemogumbe:
Do we go with python build? and align with service  libraries?

PR #286, opened 2024-06-01
Create Types for untyped nodes

### LINKED ISSUE #285: Create types for untyped nodes
To support untyped JSON (https://github.com/microsoft/kiota/issues/4385), create the following types:

`UntypedDict
`
`UntypedList
`
`UntypedStr
`
`UntypedInt
`
`UntypedBool
`
`UntypedNone
`
These types need to derive from `UntypedNode` that extends `Parsable`

rohitkrsoni:
## Overview

This PR adds Untyped nodes classes to support https://github.com/microsoft/kiota/issues/4385

## Related Issue

Fixes #285

### DIFF
--- a/kiota_abstractions/serialization/untyped_dict.py
+++ b/kiota_abstractions/serialization/untyped_dict.py
@@ -1,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedDictionary(UntypedNode):
+    """
+    Represents an untyped node with dictionary value.
+    """
+
+    def __init__(self, value: dict) -> None:
+        """
+        Creates a new instance of UntypedDictionary.
+
+        Args:
+            value (dict): The key-value pair associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> dict:
+        """
+        Gets the value associated with untyped dictionary node.
+
+        Returns:
+            The value associated with untyped dictionary node.
+        """
+        return self.__value
--- a/kiota_abstractions/serialization/untyped_list.py
+++ b/kiota_abstractions/serialization/untyped_list.py
@@ -1,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedList(UntypedNode):
+    """
+    Represents an untyped node with list value.
+    """
+
+    def __init__(self, value: list) -> None:
+        """
+        Creates a new instance of UntypedList.
+
+        Args:
+            value (bool): The list value associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> list:
+        """
+        Gets the value associated with untyped list node.
+
+        Returns:
+            The value associated with untyped list node.
+        """
+        return self.__value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added support for untyped nodes. [#285](https://github.com/microsoft/kiota-abstractions-python/issues/285)
+
+### Changed
 
 ## [1.3.3] - 2024-05-21
 
--- a/kiota_abstractions/serialization/untyped_boolean.py
+++ b/kiota_abstractions/serialization/untyped_boolean.py
@@ -1,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedBoolean(UntypedNode):
+    """
+    Represents an untyped node with boolean value.
+    """
+
+    def __init__(self, value: bool) -> None:
+        """
+        Creates a new instance of UntypedBoolean.
+
+        Args:
+            value (bool): The boolean value associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> bool:
+        """
+        Gets the value associated with untyped boolean node.
+
+        Returns:
+            The value associated with untyped boolean node.
+        """
+        return self.__value
--- a/kiota_abstractions/serialization/untyped_float.py
+++ b/kiota_abstractions/serialization/untyped_float.py
@@ -1,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedFloat(UntypedNode):
+    """
+    Represents an untyped node with float value.
+    """
+
+    def __init__(self, value: float) -> None:
+        """
+        Creates a new instance of UntypedFloat.
+
+        Args:
+            value (bool): The float value associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> float:
+        """
+        Gets the value associated with untyped float node.
+
+        Returns:
+            The value associated with untyped float node.
+        """
+        return self.__value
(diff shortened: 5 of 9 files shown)


kiota_abstractions/serialization/untyped_dict.py
```diff
@@ -0,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedDictionary(UntypedNode):
```
andrueastman, line 3:
```suggestion
class UntypedObject(UntypedNode):
```
Is there any reason to not call it `UntypedObject`?

kiota_abstractions/serialization/untyped_list.py
```diff
@@ -0,0 +1,24 @@
+from .untyped_node import UntypedNode
+
+class UntypedList(UntypedNode):
```
andrueastman, line 3:
```suggestion
class UntypedArray(UntypedNode):
```
Is there any reason to not call it UntypedArray?

kiota_abstractions/serialization/untyped_dict.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, value: dict) -> None:
+        """
+        Creates a new instance of UntypedDictionary.
+
+        Args:
+            value (dict): The key-value pair associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> dict:
```
andrueastman, line 17: is it possible to constrain the type to a dictionary of [key, UntypedNode] ?

kiota_abstractions/serialization/untyped_list.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, value: list) -> None:
+        """
+        Creates a new instance of UntypedList.
+
+        Args:
+            value (bool): The list value associated with the node.
+        """
+        self.__value = value
+
+    def get_value(self) -> list:
```
andrueastman, line 17: Is it possible to constraing this to a list of `UntypedNode`?

andrueastman (commented):
Thanks for starting this @rohitkrsoni 

Just a couple of comments here. Will you also look into writing the serialization implementation as well?

kiota_abstractions/serialization/untyped_dict.py
rohitkrsoni: Initially, I though of adding it as `UntypedObject` but in python we call it dictionary which stores key-value pairs similar to `HashMap` in `Java`. https://docs.python.org/3/tutorial/datastructures.html#dictionaries

I can change it back to `Object` if you like.

And, yes I will look into serialization implementation as well, but first I need to get my head around it.

Thanks

kiota_abstractions/serialization/untyped_dict.py
rohitkrsoni: Yes, will [str, UntypedNode] work? 
`def __init__(self, value: dict[str:UntypedNode]) -> None:`
`def get_value(self) -> dict[str:UntypedNode]:`

kiota_abstractions/serialization/untyped_list.py
rohitkrsoni: Same as above, in Python we have `list`  instead of `Arrays`, but I can change it back to Array if you like?

kiota_abstractions/serialization/untyped_list.py
rohitkrsoni: Done

rohitkrsoni:
> Thanks for starting this @rohitkrsoni
> 
> Just a couple of comments here. Will you also look into writing the serialization implementation as well?

Yes I will look into serialization implementation as well, but first I need to get my head around it.

kiota_abstractions/serialization/untyped_dict.py
andrueastman: Yeah. Constraining with the base should allow the values to be any of the derived types.

andrueastman:
> Yes I will look into serialization implementation as well, but first I need to get my head around it.

No worries. It would be great if we can get that started as well before we merge this one to avoid any unwanted regressions.

kiota_abstractions/serialization/untyped_dict.py
andrueastman: I don't have strong opinion on this one. It's better to align with what the language thinks is best. 
Maybe @shemogumbe can weigh in on the naming here..

kiota_abstractions/serialization/untyped_dict.py
rohitkrsoni: I have one more concern with regards to constraining the type of list and dictionary or any other argument type of python function. Python doesn't raises an exception if the argument passed is not of the same type. If we want to handle that, we should use try-except and assert the type. e.g. https://onlinegdb.com/u3nKT5i0M
Same goes with return type as well. These are just type hints and will not raise errors if the provided types do not match the actual values. They don't enforce type-safety.

I will implement something like that provided in the example if you like or else will just write type-hints.

thanks

kiota_abstractions/serialization/untyped_dict.py
rohitkrsoni: My view is - `arrays `and `objects ` instead of `list` and `dict` might create confusion as the codebase of the project using python client will have lists and dictionaries. 

open for suggestions.

rohitkrsoni:
> > Yes I will look into serialization implementation as well, but first I need to get my head around it.
> 
> No worries. It would be great if we can get that started as well before we merge this one to avoid any unwanted regressions.

Sure, will start that as well. I'm afraid we might need to put this on hold for some time.

kiota_abstractions/serialization/untyped_dict.py
rohitkrsoni: Ok Cool! I will implement something to raise the exception other than type hints.

baywet:
@rohitkrsoni @andrueastman just noticing this was a bit forgotten about. How can we get this finalized? (who's expecting what at this point?)

rohitkrsoni:
@baywet I was waiting for a confirmation on weather to define it as `UntypedDictionary` and `UntypedLists` or `UntypedObject` and `UntypedArrays` and then it needs review

https://github.com/microsoft/kiota-abstractions-python/pull/286#discussion_r1625671695

After that we need to work on serialization ticket.

Thanks

kiota_abstractions/serialization/untyped_dict.py
andrueastman: @rohitkrsoni Lets keep this as is for now since the it makes sense for the language to have the naming as so and then start working on the serialization bits. In the event of any feedback before we merge, we can do a renaming.

NateMerritt:
What is the current progress on adding `UntypedNode` support to Python?  Right now it seems that the generated Python models are unusable for my API until this support is added.

I have the following (simplified) OpenAPI model JSON:
```
      "Response": {
        "type": "object",
        "additionalProperties": false,
        "properties": {
          "id": {
            "type": "integer",
            "format": "int32"
          },
          "name": {
            "type": "string"
          },
          "content": {},
        }
      },
```
In the generated Python `Response` class the `content` property is missing altogether which I'm assuming is because that property doesn't have a `type` specified in the spec.  Is there a workaround for this to at least get at the original JSON send down and manually deserialize it?  In the C# and TypeScript clients the `Content` property is generated as an `UntypedNode` object so I can at least serialize it to JSON again.

Issue #302, opened 2024-07-04
Group python Kiota repositories


andrueastman:
Related to https://github.com/microsoft/kiota/issues/4636 and https://github.com/microsoft/kiota-abstractions-dotnet/issues/238

As a language specific concern, we should look into the feasibility of grouping the repos and outline if this is not feasible in this issue.

musale:
This one will be an interesting one. I have come across tooling like [pants](https://www.pantsbuild.org/) that supports Go, Java, Python, Scala, Kotlin, and Shell. It's also open source https://github.com/pantsbuild/pants

andrueastman:
Thats great!.

You can do the work as PR to this repo. Then we will rename it similar to work done for the dotnet migrations at https://github.com/microsoft/kiota-dotnet/pull/262.

andrueastman:
[note for implementation/design] See https://github.com/microsoft/kiota-abstractions-python/issues/89 for more insights.

baywet:
Whenever this is completed we should also:
- update the quickstart samples https://github.com/microsoft/kiota-samples/pull/6457
- update the quickstart documenation https://github.com/MicrosoftDocs/openapi-docs/pull/110

PR #329, opened 2024-09-10
Merge sources into one repo

andrueastman:
Fixes https://github.com/microsoft/kiota-abstractions-python/issues/302
Fixes https://github.com/microsoft/kiota-abstractions-python/issues/174

Todo

- [x] Resolved mypy issues via https://github.com/microsoft/kiota-abstractions-python/pull/331
- [x] Unify versioning across projects via https://github.com/microsoft/kiota-abstractions-python/pull/344
- [x] Sync changes for https://github.com/microsoft/kiota-serialization-json-python/pull/369 via https://github.com/microsoft/kiota-abstractions-python/pull/345
- [x] Validate workflows

PR #331, opened 2024-09-11
resolve typing errors in project after merge

andrueastman:
Partial of https://github.com/microsoft/kiota-abstractions-python/issues/302

Resolves failing mypy issues after repositories merging.

In general
- Methods return and parameter types were inconstent across the inheritance hierachy

CI for the PR at https://github.com/microsoft/kiota-abstractions-python/actions/runs/10807820407


packages/serialization/form/kiota_serialization_form/form_parse_node.py
```diff
@@ -219,9 +219,9 @@ def get_enum_value(self, enum_class: K) -> Any:
             if value not in enum_values:
                 raise Exception(f'Invalid value: {value} for enum {enum_class}.')
             result.append(enum_class(value))  # type: ignore
-        return result
+        return result  # type: ignore
 
-    def get_object_value(self, factory: ParsableFactory[U]) -> U:
+    def get_object_value(self, factory: ParsableFactory) -> Parsable:
```
baywet, line 224: This change of the return type will be problematic for auto-completion and type validation of the generated clients I believe.

packages/serialization/json/kiota_serialization_json/json_parse_node.py
```diff
@@ -206,7 +205,7 @@ def get_enum_value(self, enum_class: K) -> Optional[K]:
         except KeyError:
             return None
 
-    def get_object_value(self, factory: ParsableFactory) -> U:
+    def get_object_value(self, factory: ParsableFactory) -> Parsable:
```
baywet, line 208: same comment about the return type

packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
```diff
@@ -291,7 +291,7 @@ def write_bytes_value(self, key: Optional[str], value: bytes) -> None:
                 self.value = base64_string
 
     def write_object_value(
-        self, key: Optional[str], value: Optional[U], *additional_values_to_merge: U
+        self, key: Optional[str], value: U, *additional_values_to_merge: Optional[List[U]]
```
baywet, line 294: I believe the value should be optional in case we want to write None/null

packages/serialization/multipart/kiota_serialization_multipart/multipart_serialization_writer.py
```diff
@@ -159,7 +159,7 @@ def write_collection_of_object_values(
         raise NotImplementedError()
 
     def write_object_value(
-        self, key: Optional[str], value: Optional[U], *additional_values_to_merge: U
+        self, key: Optional[str], value: U, *additional_values_to_merge: Optional[List[U]]
```
baywet, line 162: same comment about the value being optional

packages/serialization/text/kiota_serialization_text/text_parse_node.py
```diff
@@ -170,7 +170,7 @@ def get_enum_value(self, enum_class: K) -> Optional[K]:
                 raise Exception(f'Invalid key: {camel_case_key} for enum {enum_class}.')
         return None
 
-    def get_object_value(self, factory: ParsableFactory) -> U:
+    def get_object_value(self, factory: ParsableFactory) -> Parsable:
```
baywet, line 173: same comment about the return type

packages/serialization/text/kiota_serialization_text/text_serialization_writer.py
```diff
@@ -158,7 +158,7 @@ def write_bytes_value(self, key: Optional[str], value: bytes) -> None:
             self.write_str_value(key, base64_string)
 
     def write_object_value(
-        self, key: Optional[str], value: Optional[U], *additional_values_to_merge: U
+        self, key: Optional[str], value: U, *additional_values_to_merge: Optional[List[U]]
```
baywet, line 161: same comment about the value being optional

packages/serialization/form/kiota_serialization_form/form_serialization_writer.py
```diff
@@ -178,7 +177,7 @@ def write_collection_of_object_values(
         raise Exception("Form serialization does not support collections.")
 
     def write_object_value(
-        self, key: Optional[str], value: Optional[U], *additional_values_to_merge: U
+        self, key: Optional[str], value: U, *additional_values_to_merge: Optional[List[U]]
```
baywet, line 180: same comment about the value being optional

packages/abstractions/kiota_abstractions/serialization/serialization_writer.py
```diff
@@ -157,7 +157,7 @@ def write_bytes_value(self, key: Optional[str], value: bytes) -> None:
 
     @abstractmethod
     def write_object_value(
-        self, key: Optional[str], value: U, additional_values_to_merge: Optional[List[U]]
+        self, key: Optional[str], value: U, *additional_values_to_merge: Optional[List[U]]
```
baywet, line 160: I believe the value should be optional in case we need to write None/null. Fixing it here and running the linting should probably address all my other comments in implementations.

packages/abstractions/kiota_abstractions/serialization/parse_node.py
```diff
@@ -118,7 +118,7 @@ def get_time_value(self) -> Optional[time]:
         pass
 
     @abstractmethod
-    def get_collection_of_primitive_values(self) -> List[T]:
+    def get_collection_of_primitive_values(self, primitive_type) -> Optional[List[T]]:
         """Gets the collection of primitive values of the node
```
baywet, line 122: doc comment is missing the args

packages/abstractions/kiota_abstractions/serialization/parse_node.py
```diff
@@ -118,7 +118,7 @@ def get_time_value(self) -> Optional[time]:
         pass
 
     @abstractmethod
-    def get_collection_of_primitive_values(self) -> List[T]:
+    def get_collection_of_primitive_values(self, primitive_type) -> Optional[List[T]]:
```
baywet, line 121: If we're changing the return type to Optional, we should do so as well in the other collection methods.

packages/abstractions/kiota_abstractions/serialization/parse_node.py
```diff
@@ -146,7 +146,7 @@ def get_collection_of_enum_values(self) -> List[K]:
         pass
 
     @abstractmethod
-    def get_enum_value(self) -> Enum:
+    def get_enum_value(self, enum_class: K) -> Optional[K]:
```
baywet, line 149: leaving the comment here, but the argument for get object value factory should be generic so the return type is also generic

baywet (changes requested):
overall this is great as it uncovers a bunch of defects in the type declarations.
I'm a bit worried about the blast radius for anybody relying on static type checking? (I know not everyone does that in Python)

packages/abstractions/kiota_abstractions/serialization/parse_node.py
andrueastman: Changed this back as non of the implementations seem to return none at the moment.

packages/serialization/json/kiota_serialization_json/json_parse_node.py
```diff
@@ -140,8 +139,10 @@ def get_time_value(self) -> Optional[time]:
                 return datetime_obj
         return None
 
-    def get_collection_of_primitive_values(self, primitive_type: Any) -> Optional[List[T]]:
+    def get_collection_of_primitive_values(self, primitive_type: Any) -> List[T]:
```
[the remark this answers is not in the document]
andrueastman: I changed this back as none of the implementations do return a none in any case. 
Unless the suggestion is to have this for possible alternative implementations?

packages/http/httpx/kiota_http/httpx_request_adapter.py
```diff
@@ -152,8 +152,8 @@ def _start_local_tracing_span(self, name: str, parent_span: trace.Span) -> trace
     async def send_async(
         self,
         request_info: RequestInformation,
-        parsable_factory: ParsableFactory,
-        error_map: Dict[str, ParsableFactory],
+        parsable_factory: ParsableFactory[ModelType],
+        error_map: Dict[str, Optional[ParsableFactory]],
```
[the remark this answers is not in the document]
andrueastman: That makes sense now. Will action and resolve..

packages/serialization/json/kiota_serialization_json/json_parse_node.py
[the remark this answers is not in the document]
baywet: this is where my Python knowledge is showing its limits: Optional is effectively the equivalent of "?" in TS/C# to me: if it's here, it's clear you should expect null/none as a result, if it's not, it's not impossible to get one, but generally unexcepted.

How close is that perception to what it actually means?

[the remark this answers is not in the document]
andrueastman: You're quite right. After taking a look at the generated library and seeing the generated code generates the properties as optional. I've changed this across the `get_collection_of_XXX` methods.

andrueastman:
>yes, having a smoke client with static type checking would help I believe.

@baywet Adding the smoke client uncovered a number of new issues some of which seem to need a bit more refactoring (generation as well) and would better be dealt with once we have achieved the goal of merging the source. 
I've created https://github.com/microsoft/kiota-abstractions-python/issues/333 for the issue with the ParsableFactory and ideally an expected outcome of that would be to have the generated client added as part of the work.

For now, I believe we can focus on the validation of the typings across the mono repo then add that once https://github.com/microsoft/kiota-abstractions-python/pull/329 is merged in. 

Any qualms? Thoughts?

andrueastman:
Created https://github.com/microsoft/kiota-abstractions-python/issues/333 to resolve parsable generation typing issues.

andrueastman:
@baywet 
https://github.com/microsoft/kiota/pull/5453 should resolve generation gaps while https://github.com/microsoft/kiota-abstractions-python/pull/340 should add a test project and fix gaps in the libs. 

Once those are closed, and this is good to go, I believe we can merge this one and now simply sort out the workflows to merge the repositories.

baywet:
Thanks for the follow up here.
What I'd like a clear answer on is the following: are these changes going to break builds for people with existing clients if they don't refresh the generated code?

andrueastman:
The short answer is not really. Builds are already broken(for users who try to consume the typing info of our packages) as we were really not publishing the typing information. 

The error/warning that shows up for current packages looks something like this.
```shell
Skipping analyzing "kiota_serialization_form.form_parse_node_factory": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

According to [PEP 561](https://typing.readthedocs.io/en/latest/spec/distributing.html#packaging-type-information), for the type information to be consumed downstream, a package author either
- MUST add a marker file named py.typed to their package supporting typing
- Publish a separate stub only package with the type information

If you take a look at the sources, none of them specify a `py.typed` marker(and we definitely don't publish a separate stub package).
- [Abstractions](https://github.com/microsoft/kiota-abstractions-python/tree/main/kiota_abstractions)
- [Auth](https://github.com/microsoft/kiota-authentication-azure-python/tree/main/kiota_authentication_azure)
- [Http](https://github.com/microsoft/kiota-http-python/tree/main/kiota_http)
- [Serialization json](https://github.com/microsoft/kiota-serialization-json-python/tree/main/kiota_serialization_json)
- [Serialization text](https://github.com/microsoft/kiota-serialization-text-python/tree/main/kiota_serialization_text)
- [Serialization multipart](https://github.com/microsoft/kiota-serialization-multipart-python/tree/main/kiota_serialization_multipart)

What this means then is that for the consumer, currently, they would either
- Write their own stub files
- Suppress the error of missing types by passing the `ignore_missing_imports = True` to the checker so that 3rd party packages that are not PEP 561 compliant are ignored.

From the sources, it looks like we have been also suppressing type validation across projects and this explains why the typing issues discovered here were not surfacing across projects. See examples below (but this is across all projects)
- https://github.com/microsoft/kiota-serialization-multipart-python/blob/211057c8c0d9e10283009eadde63a16d374060e5/pyproject.toml#L36
- https://github.com/microsoft/kiota-abstractions-python/blob/c2d52b504f0a283e906bd6f756a21ab52b406fa1/pyproject.toml#L38
- https://github.com/microsoft/kiota-serialization-json-python/blob/51189ef48a8e298c691025bb6d426b4b0d40f0c7/pyproject.toml#L36

As we are moving to a mono repo structure, the packages do not see each other as 3rd party anymore as they are co-located and thus the suppression won't hide these type errors anymore. So, if we wish to keep the type checking we would need to fix them. 

The end result of this process then will be
- Our packages will be [PEP 561](https://peps.python.org/pep-0561/) compliant(as we have added the py.typed markers in this PR and https://github.com/microsoft/kiota-abstractions-python/pull/329) so the typing information will consumed downstream.(no need for supressions when consuming packages)
- Typing information will be correct (validated with the generated client)

packages/abstractions/kiota_abstractions/serialization/parse_node.py
```diff
@@ -137,7 +138,7 @@ def get_collection_of_object_values(self, factory: ParsableFactory) -> List[U]:
         pass
 
     @abstractmethod
-    def get_collection_of_enum_values(self) -> List[K]:
+    def get_collection_of_enum_values(self, enum_class: K) -> List[Optional[K]]:
```
baywet, line 141:
```suggestion
    def get_collection_of_enum_values(self, enum_class: K) -> Optional[List[K]]:
```

packages/abstractions/kiota_abstractions/serialization/parse_node.py
```diff
@@ -155,17 +156,17 @@ def get_enum_value(self) -> Enum:
         pass
 
     @abstractmethod
-    def get_object_value(self, factory: ParsableFactory) -> Parsable:
+    def get_object_value(self, factory: ParsableFactory[U]) -> U:
```
baywet, line 159: should U be optional here?

baywet: sorry, I didn't mean for the argument, but for the return type.

baywet:
@andrueastman as a side note, this PR will also need to be replicated here https://github.com/microsoft/kiota-serialization-json-python/pull/369

packages/abstractions/kiota_abstractions/serialization/parse_node.py
andrueastman: I believe we're talking about the same thing. 
As `get_object_value` will call the `ParsableFactory` (which essentially calls `create_from_discriminator_value`). Since this never returns an optional, we won't get an optional returned as `get_object_value` simply populates the properties of the object returned by the factory and returns the modified object.

andrueastman:
> ft/kiota-serialization-json-python#369

Thanks for highlighting this @baywet 

I've added this as a task in the bigger PR at https://github.com/microsoft/kiota-abstractions-python/pull/329 so that we try close this and not mix too many concerns here. I'll create a separate PR into the bigger PR once this one is closed.

packages/abstractions/kiota_abstractions/serialization/parse_node.py
baywet: It seems that we have disparities between Java and dotnet...

https://github.com/microsoft/kiota-dotnet/blob/3b1f96c2aa64a09792133838e4e0184038fa985f/src/abstractions/serialization/IParseNode.cs#L132

https://github.com/microsoft/kiota-java/blob/ce707fc2e11f3b2add49880365b47b87a5c2b62a/components/abstractions/src/main/java/com/microsoft/kiota/serialization/ParseNode.java#L160

But after thinking about it some more, I believe you're correct, GetObjectValue will only be called if the property is included and non null.

Good to resolve this one, I'll just create an issue in kiota dotnet.

baywet (approved):

PR #381, opened 2024-10-16
fix: ensure all packages are released together.

andrueastman:
Looks like according to https://github.com/googleapis/release-please/pull/1749 for the `linked-versions` plugin to align the versions, the component will need to be included in the tag. 

This will change the tag format in the repo from `v1.2.0` to be `microsoft-kiota-abstractions-v1.2.0` (therefore publishing a tag for each library similar to [kiota-typescript](https://github.com/microsoft/kiota-typescript/tags))

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,7 @@ name: Publish package to PyPI
 on:
   push:
     tags:
-      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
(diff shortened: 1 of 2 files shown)


baywet (approved):

baywet:
@andrueastman the v prefix is missing now/ this filter needs to be updated https://github.com/microsoft/kiota-python/blob/2d532046296baf5556b5d2b4c0118b3920472ddc/.github/workflows/publish.yml#L19

andrueastman:
Thanks for catching that. https://github.com/microsoft/kiota-python/pull/383 should allow us to validate if everything will work well including future version bumps.

PR #385, opened 2024-10-16
remove unused code/imports as well as some reformatting

Acurisu:
## Overview

This PR is not needed per se, just some linting related stuff that I noticed while working on something else.
Increases PyLint score slightly.


andrueastman (approved):
Thanks for the contribution here @Acurisu

PR #386, opened 2024-10-16
correctly serialize None in a list

Acurisu:
## Overview

We currently have two issues:
1. We don't support `None` inside lists.
2. We convert `None` to `'null'` when there is no key. Simply fixing issue 1 would result in `['null', 'null']` instead of `[null, null]`.

The first issue arises when fetching, for example, all graph site pages of a site and attempting to serialize them. A specific case involves expanding `canvasLayout` and receiving a page without a `canvasLayout`, which results in `pinnedItems` containing `[None, None, None]`.

Although whether this response is correct is another question, the serialization should not raise an exception in this scenario but should handle it appropriately.

## Testing Instructions

I have added relevant tests. Simply run `pytest` to verify the fix. Running the same tests without my changes will result in an exception.

### DIFF
--- a/packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
+++ b/packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
@@ -14,6 +14,7 @@ T = TypeVar("T")
 U = TypeVar("U", bound=Parsable)
 K = TypeVar("K", bound=Enum)
 PRIMITIVE_TYPES = [bool, str, int, float, UUID, datetime, timedelta, date, time, bytes, Enum]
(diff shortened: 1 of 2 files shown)


andrueastman:
Thanks for opening this @Acurisu 

If we were to include this change, it would however not trigger a release. Any chance you can include/modify commits to follow the conventions here 
https://github.com/microsoft/kiota-python/blob/main/CONTRIBUTING.md#commit-message-format

Acurisu:
Done. I hope this works out.

andrueastman (approved):
Thanks again for the contribution @Acurisu

PR #391, opened 2024-10-16
fix/outdated deps

baywet:
- **fix: version requirement for httpx**
- **fix: dependabot configuration for new project structure**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,15 @@
 version: 2
 updates:
 - package-ecosystem: pip
-  directory: "/"
+  directories: 
+    - "/packages/abstractions"
+    - "/packages/serialization/form"
+    - "/packages/serialization/json"
+    - "/packages/serialization/text"
+    - "/packages/serialization/multipart"
+    - "/packages/authentication/azure"
+    - "/packages/http/httpx"
+    - "/packages/tests/validation"
   schedule:
     interval: daily
   open-pull-requests-limit: 10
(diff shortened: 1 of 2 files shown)


andrueastman:
I believe there's a conflict with https://github.com/microsoft/kiota-python/pull/387

baywet:
@andrueastman yes, I believe we identified the problem at about the same time. We have a bunch of outdated dependencies in the toml files. (not sure whether it's a regression, old repos used the requirements file)

Also from looking [at the dependabot documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#pip-and-pip-compile)
And the [toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec) it's possible our projects are behind (no project section, missing dependencies, using poetry dependencies, etc...) what do you think?

andrueastman:
Just checking in here. Any chance this is still an issue after the dependabot PRs? 
https://github.com/microsoft/kiota-python/pull/389

I believe dependabot should be able to handle the poetry deps. as it is listed in below. 
https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/dependency-graph-supported-package-ecosystems#supported-package-ecosystems

I believe the discrepancy was caused by incorectly pulling the version range modifiers from the original projects (using `^` vs `>=`) but dependabot seems to have fixed that now....

baywet:
I wasn't sure why we were not getting any dependabot PRs on the toml. I thought it was because of the file "format" but now that we've received a PR I think here is what happened:
1. pre grouping, since we had requirements files, dependabot was updating those and the toml fell behind.
2. after grouping, the directories were wrong, and we removed the requirement files. So we weren't getting dependabot PRs, and the tomls were migrated with "behind dependencies" from their original repos
3. now that the directories are fixes we're getting dependbot PRs.

If this is your understanding as well. I don't think we need to take further action at this point, At least not for the dependabot aspects. Maybe we'll want to mordernize our toml files for other reasons, but nothing pressing as far as I can tell.

PR #393, opened 2024-10-18
fix: fixes typing errors from integration tests for optional parameters

andrueastman:
Fixes errors related to https://github.com/microsoft/kiota/pull/5617

### DIFF
--- a/tests/validation/pyproject.toml
+++ b/tests/validation/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.poetry]
 name = "validation-client"
 version = "1.3.4"
+package-mode = false
 description = "Core abstractions for kiota generated libraries in Python"
 authors = ["Microsoft <graphtooling+python@microsoft.com>"]
 license = "MIT"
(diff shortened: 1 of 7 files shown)


tests/validation/pyproject.toml
```diff
@@ -5,6 +5,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.poetry]
 name = "validation-client"
 version = "1.3.4"
+package-mode = false
```
baywet, line 8: this is something I'm not clear on. It seems that between this repo and the Microsoft Graph ones, we might have different ways to package our packages.
That being said, if we change this here, shouldn't it be consistent across all packages on this repo?
https://python-poetry.org/docs/basic-usage/#operating-modes

andrueastman: This is specifically for the `validation` test project. Since its not being package for release, we don't want to have the running of any commands to be interpreted for package mode. We're really only interested in dependency management in this scenario...

baywet: ah sorry, I did have my tea yet. Ignore me.

baywet (approved):

PR #417, opened 2024-12-02
Fixes Retry Handler delay validation, retry count validation and exponential backoff

Ndiritu:
- Ensure calculated retry delay validated against correct maximum delay of 180 secs
-  Fixes retry handler exponential back-off to consider the delay specified in the retry handler option
- Ensures retry count is incremented based on value in retry-attempt header to prevent exceeding configured max retries

closes https://github.com/microsoft/kiota-python/issues/409

### DIFF
--- a/packages/http/httpx/kiota_http/middleware/retry_handler.py
+++ b/packages/http/httpx/kiota_http/middleware/retry_handler.py
@@ -88,7 +88,7 @@ class RetryHandler(BaseMiddleware):
             retry_valid = self.check_retry_valid(retry_count, current_options)
 
             # Get the delay time between retries
-            delay = self.get_delay_time(retry_count, response)
+            delay = self.get_delay_time(retry_count, response, current_options.max_delay)
 
             # Check if the request needs to be retried based on the response method
             # and status code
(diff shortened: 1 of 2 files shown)


packages/http/httpx/kiota_http/middleware/retry_handler.py
```diff
@@ -71,43 +73,41 @@ async def send(self, request: httpx.Request, transport: httpx.AsyncBaseTransport
...
 
             # Get the delay time between retries
-            delay = self.get_delay_time(retry_count, response)
+            delay = self.get_delay_time(retry_count, response, current_options.max_delay)
 
             # Check if the request needs to be retried based on the response method
             # and status code
             should_retry = self.should_retry(request, current_options, response)
-            if all([should_retry, retry_valid, delay < max_delay]):
+            if all([should_retry, retry_valid, delay < RetryHandlerOption.MAX_DELAY]):
```
baywet, line 99: shouldn't we take the smallest between the constant max delay and the option max delay here?

Ndiritu: I think we named the optiona variable name wrongly and unfortunately it's exposed publicly so we can't change it.

From my understanding, in the options, it was supposed to be the minimum delay to wait before retrying. Otherwise we do the exponential backoff that should consider the current retry attempt & delay but should not exceed the max constant

baywet (approved):

PR #421, opened 2024-12-03
fix: Regex extracting claims (addressing #420)

### LINKED ISSUE #420: http - incorrect regex causing claim parsing to fail
Incorrect regex in `packages/http/httpx/kiota_http/httpx_request_adapter.py` causes downstream impact decoding the base-64 encoded claim in reponse.

**Environment:**
Python Version: 3.13.0
kiota-python version: 1.6.3 
OS: Windows 11 Version 10.0.22621 Build 22621

**Stack trace:**
Image

**Describe the bug:**
Calling `get_authoritzation_token`  to retrieve a bearer token using the `AzureIdentityAccessTokenProvider` class generates a `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x83 in position 109: invalid start byte` exception.

The issue stems from the incorrect regex used in the `retry_cae_response_if_required` function in file `packages/http/httpx/kiota_http/httpx_request_adapter.py`. The following regex attempts to extract the base-64 encoded claims included in the value of the `WWW-Authenticate` header:

`claims="(.+)"`

However, the `.+` greedily matches all characters up to the final `"` present in the `WWW-Authenticate` header. If the value contains any other keys (e.g. `Pop realm`, `authorization_uri`, `client_id` etc.) they are also captured.

For example, suppose the value of the `WWW-Authenticate` header is:
`claims="eyJhYsdgsXSd", PoP realm="", authorization_uri="https://login.microsoftonline.com/common/oauth2/authorize", nonce="eyJ0e"`

The expected behavior of the regex is to match `claims="eyJhYsdgsXSd"`; however, it instead matches the full string up to an including the final `...J0e"`. The function subsequent splitting on the string `="` means the "base-64" encoded claims passed to subsequent functions would be the string `eyJhYsdgsXSd", PoP realm`. As this is an invalid base-64 encoded string, when it is decoded it raises a `UnicodeDecodeError` exception.

**Suggested fix:**
Replace the regex with `claims="([^"]+)"` and use the first matching group to extract the base-64 encoded value. I will be raising a PR shortly.

**Expected behavior:**
A bearer token would be returned.

Mrfence97:
## Overview

Updates the regex extracting the base-64 encoded claims in the WWW-Authenticate header. More details provided in issue #420.

## Related Issue

Fixes #420

### DIFF
--- a/packages/http/httpx/kiota_http/httpx_request_adapter.py
+++ b/packages/http/httpx/kiota_http/httpx_request_adapter.py
@@ -564,10 +564,10 @@ class HttpxRequestAdapter(RequestAdapter):
             if auth_header_value.casefold().startswith(
                 self.BEARER_AUTHENTICATION_SCHEME.casefold()
             ):
-                claims_match = re.search('claims="(.+)"', auth_header_value)
+                claims_match = re.search('claims="([^"]+)"', auth_header_value)
                 if not claims_match:
                     raise ValueError("Unable to parse claims from response")
-                response_claims = claims_match.group().split('="')[1]
+                response_claims = claims_match.group(1)
                 parent_span.add_event(AUTHENTICATE_CHALLENGED_EVENT_KEY)
                 parent_span.set_attribute("http.retry_count", 1)
                 return await self.get_http_response_message(
--- a/packages/http/httpx/tests/test_httpx_request_adapter.py
+++ b/packages/http/httpx/tests/test_httpx_request_adapter.py
@@ -394,7 +394,7 @@ async def test_retries_on_cae_failure(
             {
                 "claims": (
                     "eyJhY2Nlc3NfdG9rZW4iOnsibmJmIjp7ImVzc2VudGlhbCI6dHJ1ZSwgInZhbH"
-                    "VlIjoiMTYwNDEwNjY1MSJ9fX0"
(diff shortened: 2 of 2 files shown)


baywet (dismissed):
Thanks for the contribution!

musale (dismissed):

musale:
@Mrfence97 could you please check some tests are failing.

Ndiritu (commented):
Adding a unit test for this would be great.

Mrfence97:
> @Mrfence97 could you please check some tests are failing.

The failing test was due to a bug in the test. In particular, there is a bug in the assertion of the `test_retries_on_cae_failure` test in `http/httpx/tests/test_httpx_request_adapter.py`. The mock claim entered in `mock_cae_failure_response` in `http/httpx/tests/conftest.py` ends with a `=`. It is unclear how this test ever passed previously!

I've added a commit to fix the test.

Mrfence97:
> Adding a unit test for this would be great.

Unfortunately I don't know your testing framework well enough to implement something myself, but hopefully somebody at Microsoft can do it!

Mrfence97:

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

baywet (approved):

andrueastman (approved):

PR #431, opened 2025-01-02
fix: opentelemetry warning

### LINKED ISSUE #430: warning: Invalid type NoneType for attribute 'http.port' value
When logging request information, the `http.port` should only be set if `url.port` is not `None`.
https://github.com/microsoft/kiota-python/blob/ac46dd6f566581364fe470262508ef71b5c5cc53/packages/http/httpx/kiota_http/httpx_request_adapter.py#L604
Currently, if the `http.port` is `None`, `opentelemetry` produces the following warning:
```
Invalid type NoneType for attribute 'http.port' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```
based on this line of code:
https://github.com/open-telemetry/opentelemetry-python/blob/29aad2efa30a35757ea78d90e4d3fd517bc00468/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L100

See https://github.com/pydantic/logfire/issues/749 for more details.

mplemay:
## Overview

Fixed the `opentelemetry` warning described in the corresponding github issue.

## Related Issue

fixes #430 

### Demo

N/A

### Notes

N/A

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/packages/http/httpx/kiota_http/httpx_request_adapter.py
+++ b/packages/http/httpx/kiota_http/httpx_request_adapter.py
@@ -601,11 +601,13 @@ class HttpxRequestAdapter(RequestAdapter):
 
         otel_attributes = {
             HTTP_REQUEST_METHOD: method.value,
-            "http.port": url.port,
             SERVER_ADDRESS: url.hostname,
             URL_SCHEME: url.scheme,
             "url.uri_template": request_info.url_template,
         }
+
+        if url.port is not None:
+            otel_attributes["http.port"] = url.port
 
         if self.observability_options.include_euii_attributes:
             otel_attributes.update({URL_FULL: url.geturl()})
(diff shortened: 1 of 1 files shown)


baywet (approved):
Thanks for the contribution!

packages/http/httpx/kiota_http/httpx_request_adapter.py
```diff
@@ -601,12 +601,14 @@ def get_request_from_request_information(
...
         otel_attributes = {
             HTTP_REQUEST_METHOD: method.value,
-            "http.port": url.port,
             SERVER_ADDRESS: url.hostname,
             URL_SCHEME: url.scheme,
             "url.uri_template": request_info.url_template,
         }
 
+        if url.port is not None:
+            otel_attributes["http.port"] = url.port
```
baywet, line 610: can you call a string conversion here please so we maintain the 3.8 compatibility?

mplemay: Done!

baywet (approved):

PR #433, opened 2025-01-07
feat: drop support for python 3.8

andrueastman:
Fixes https://github.com/microsoft/kiota-python/issues/349

This PR drops support for python 3.8 in the kiota libraries.

In summary, the changes include
- Dropping using of collection type hint aliases such as Dict, Set and List from typing. 
- Updates projects requirements to include python 3.9+ and drop 3.8
- Updates validation client to use non deprecated hints via https://github.com/microsoft/kiota-python/pull/434

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       max-parallel: 10
       matrix:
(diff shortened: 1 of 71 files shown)


baywet (approved):
Thanks for also taking the time to clean up the imports of typings to BCL, I had not anticipated that.

PR #450, opened 2025-01-14
Remove pendulum dependency

### LINKED ISSUE #448: json/form/text - remove pendulum dependency
related #348

Pendulum is a fairly heavy dependency (requires cargo at install???), which maintenance over time doesn't see guaranteed (author fatigue).

We have only 3 (non-test) files making references to this dependency, with a total of 12 references to parsing functions.
https://github.com/search?q=repo%3Amicrosoft%2Fkiota-python+pendulum+language%3APython+NOT+path%3A%2F%5Epackages%5C%2Fserialization%5C%2Fjson%5C%2Ftests%5C%2Funit%5C%2F%2F+NOT+path%3A%2F%5Epackages%5C%2Fserialization%5C%2Fform%5C%2Ftests%5C%2Funit%5C%2F%2F&type=code&l=Python

Except for time delta, all those parsing functions have equivalents in the main python BCL.

https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat
https://docs.python.org/3/library/datetime.html#datetime.time.fromisoformat
https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat

Here are the steps to take:
- [ ] implement a static function parseTimeDeltaFromIsoFormat in abstractions [use the duration ABNF from RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#appendix-A)
- [ ] replace all calls to pendulum by the equivalent BCL/abstractions methods
- [ ] remove the dependency from the packages.

svrooij:
Fixes #448

Remove the `pendulum` dependency and replace it with equivalent BCL/abstractions methods.

* **Add `parseTimeDeltaFromIsoFormat` function**:
  - Implement a static function `parseTimeDeltaFromIsoFormat` in `packages/abstractions/kiota_abstractions/utils.py` to parse ISO 8601 duration strings.
  - Add import for `re` module.

* **Replace `pendulum` calls**:
  - Replace `pendulum.parse` calls with equivalent BCL/abstractions methods in `packages/serialization/form/kiota_serialization_form/form_parse_node.py`.
  - Replace `pendulum.parse` calls with equivalent BCL/abstractions methods in `packages/serialization/json/kiota_serialization_json/json_parse_node.py`.

* **Remove `pendulum` dependency**:
  - Remove `pendulum` dependency from `packages/serialization/form/pyproject.toml`.
  - Remove `pendulum` dependency from `packages/serialization/json/pyproject.toml`.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/kiota-python/pull/450?shareId=fba017d2-eb80-411d-80e2-a5d8e31fa519).


svrooij:
To be totally clear, I'm not a python developer this is all copilot 😄

packages/abstractions/kiota_abstractions/utils.py
```diff
@@ -37,3 +38,34 @@ def lazy_import(name):
...
+        r'(?:(\d+)H)?'  # hours
+        r'(?:(\d+)M)?'  # minutes
+        r'(?:(\d+)S)?)?'  # seconds
+    )
+    match = pattern.fullmatch(duration_str)
+    if not match:
+        raise ValueError(f"Invalid ISO 8601 duration string: {duration_str}")
+
+    years, months, days, hours, minutes, seconds = match.groups()
+    return timedelta(
```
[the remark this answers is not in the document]
svrooij: > The timedelta class in Python's datetime module does not enforce faulty combinations like months and weeks. Instead, it computes the expected value based on the provided arguments. The timedelta class only supports days, seconds, and microseconds directly, but you can specify weeks, hours, minutes, and milliseconds, which are converted into the appropriate number of days, seconds, and microseconds.

So internally it just computes an expected value if you combine weeks with days (or other values). Maybe stay as close as possible to the python library?

packages/abstractions/kiota_abstractions/utils.py
```diff
@@ -37,3 +38,34 @@ def lazy_import(name):
...
+    Args:
+        duration_str (str): The ISO 8601 duration string.
+
+    Returns:
+        timedelta: The parsed timedelta object.
+    """
+    pattern = re.compile(
+        r'P'  # starts with 'P'
+        r'(?:(\d+)Y)?'  # years
+        r'(?:(\d+)M)?'  # months
```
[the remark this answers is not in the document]
svrooij: Added

packages/serialization/json/tests/unit/test_json_parse_node.py
```diff
@@ -161,8 +161,8 @@ def test_get_anythin_does_not_convert_any_length_numeric_chars_to_datetime():
 def test_get_anythin_does_convert_date_string_to_datetime():
     parse_node = JsonParseNode("2023-10-05T14:48:00.000Z")
     result = parse_node.try_get_anything("2023-10-05T14:48:00.000Z")
-    assert isinstance(result, pendulum.DateTime)
-    assert result == pendulum.parse("2023-10-05T14:48:00.000Z")
+    assert isinstance(result, datetime)
+    assert result == datetime.fromisoformat("2023-10-05T14:48:00.000Z")
```
baywet, line 165: imports L4 and 7 need to be removed

packages/abstractions/kiota_abstractions/utils.py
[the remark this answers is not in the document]
baywet: that'd diverge from the other languages. Also we should check what pendulum was doing with those scenarios, as we might be introducing a regression

svrooij:
@baywet is there a command to apply the wanted code formatting? I see it's failing at that.

baywet:
@svrooij should be something like `yapf -ir .`

baywet:
@svrooij a few files were forgotten about:
- [ ] packages\serialization\form\tests\unit\test_form_serialization_writer.py
- [ ] packages\serialization\json\kiota_serialization_json\json_serialization_writer.py
- [ ] packages\serialization\json\tests\unit\test_json_parse_node.py
- [ ] packages/serialization/multipart/pyproject.toml
- [ ] packages\serialization\json\tests\unit\test_json_serialization_writer.py

baywet:
@svrooij there are additional CI issues, would you mind having a look at it please?

svrooij:
This is related to #453 being able to run the tests locally would greatly reduce the developer feedback flow.

baywet:
@svrooij I have replied to the issue you've created about running unit tests.
I've also corrected a lot of them to help you get closer to completion.
I believe most of the ones still failing at this point are related to how durations are parsed.
Would you mind taking over to get us across the finish line here please?

svrooij:
@baywet I'll try to get it over the finish line, tomorrow.

svrooij:
Got the tests working in a dev container. Down to the last issue.

```plain
text = 'T12H59M45S'

    def parse_timedelta_from_iso_format(text: str) -> timedelta:
        """Parses a ISO8601 duration string into a timedelta object."""
    
        m = _ISO8601_DURATION_PATTERN.match(text)
        if not m:
>           raise ValueError(f"Invalid ISO8601 duration string: {text}")
E           ValueError: Invalid ISO8601 duration string: T12H59M45S
```

It seems that pendulum supported "invalid" string, if I purely check the ISO8601 standard it should be `PT12H59M45S` (with the `P` in front).

To make it not stop working my suggestion is that I make the P optional.

baywet:
> To make it not stop working my suggestion is that I make the P optional.

Ah no, good catch, it's a typo on my end, please add the P, it's mandatory according to the ABNF.

svrooij:
> > To make it not stop working my suggestion is that I make the P optional.
> 
> Ah no, good catch, it's a typo on my end, please add the P, it's mandatory according to the ABNF.

Want me to remove?

baywet:
> > > To make it not stop working my suggestion is that I make the P optional.
> > 
> > 
> > Ah no, good catch, it's a typo on my end, please add the P, it's mandatory according to the ABNF.
> 
> Want me to remove?

Just add the "P" for the unit tests where it's missing, and make the P mandatory in the regex again :)
Thank you and sorry for the confusion!

packages/abstractions/kiota_abstractions/date_utils.py
```diff
@@ -0,0 +1,76 @@
+import re
+from datetime import timedelta
+
+
+_ISO8601_DURATION_PATTERN = re.compile(
+    "^P"  # Duration P indicator
+    # Weeks
+    "(?P<w>"
+    r"    (?P<weeks>\d+(?:[.,]\d+)?W)"
```
baywet, line 9: are those `?P` a special syntax in Python regex for group naming? or are we looking for the mandatory P character in all sub-segments?

baywet (approved):

packages/abstractions/kiota_abstractions/date_utils.py
svrooij: Yes ?P is python regex for, this is a named group

packages/serialization/text/tests/unit/test_text_parse_node.py
```diff
@@ -66,21 +66,21 @@ def test_get_datetime_value():
...
-    parse_node = TextParseNode('2022-01-27T12:59:45.596117')
+    parse_node = TextParseNode('12:59:45.596117')
     result = parse_node.get_time_value()
     assert isinstance(result, time)
     assert str(result) == '12:59:45.596117'
 
 
 def test_get_timedelta_value():
-    parse_node = TextParseNode('2022-01-27T12:59:45.596117')
+    parse_node = TextParseNode('PT12H59M45S')
```
danieleds, line 83: what's the reason for these changes in the tests?

danieleds:
The tests are failing because `time.fromisoformat` accepted a limited set of formats in Python 3.9 and 3.10.

In particular, the tests specify

```
    "startWorkTime=08:00:00.0000000&"
    "endWorkTime=17:00:00.0000000&"
```

but on older python, the fractional second part is only allowed to have either 3 or 6 digits, and we specify 7.

An option would be, if we are on an older python and there is a fractional part ("`.`"), then we cut it to 6 if the input string is longer than 6, or pad it with zeros to 6 if it is shorter

baywet:
thanks for chiming in @danieleds !!

to achieve this, would you add a condition [like this one](https://stackoverflow.com/a/20624861/3808675) in the code?

danieleds:
@baywet would love to, but I don't think I have push permissions on this branch

danieleds:
Oh I see. Yes, that's exactly what I had in mind. Something like, in form_parse_node.py and similar files:

```py
if sys.version_info[:3] <= (3,10):
    fixed_time = re.sub(r'(\d[.,])(\d+)', lambda x: x.group(1) + x.group(2).ljust(6,'0')[:6], self._node)
    return time.fromisoformat(fixed_time)
else:
    return time.fromisoformat(self._node)
```

(untested)

svrooij:
> Oh I see. Yes, that's exactly what I had in mind. Something like, in form_parse_node.py and similar files:
> 
> ```python
> if sys.version_info[:3] <= (3,10):
>     fixed_time = re.sub(r'(\d[.,])(\d+)', lambda x: x.group(1) + x.group(2).ljust(6,'0')[:6], self._node)
>     return time.fromisoformat(fixed_time)
> else:
>     return time.fromisoformat(self._node)
> ```
> 
> (untested)

Thanks @danieleds for your input.
Since that code is used multiple times, I added it to the date_utls file.

https://github.com/microsoft/kiota-python/pull/450/commits/8aaeff1e446b3503af46ca6a51261871914fa77a#diff-8acdec4a0f6630bb416ac06117c85639c6656c1f65ef249c859dad74bc82c0ccR82-R98

baywet:
@svrooij I did have a question: when I approve the CI and it fails, do you receive notifications on your end or not? I don't want to double down in case you're already receiving those :)

svrooij:
@baywet I only get notifications from reactions (yours and sonarqube)

baywet:
@svrooij Thanks for confirming. That's what I suspected since I get the notifications when the CI fails. So I'll keep notifying you here if that works for you.

And yes, the CI failed again :)

svrooij:
Going to download a python 3.9 dev container and run the tests there.

danieleds:
@svrooij looks like in the regex we also need to replace the "`,`" with "`.`" for older python.

Also I realized my python version check was slightly wrong, I suggest using the following instead: `if sys.version_info[:3] < (3,11):` (so that all patch versions of 3.10 do match too)

svrooij:
@danieleds and strip of the `Z`?

```
FAILED tests/unit/test_json_parse_node.py::test_get_collection_of_object_values - ValueError: Invalid isoformat string: '2021-07-29T03:07:25Z'
```

danieleds:
@svrooij yes I think that's the issue. Instead of stripping the "Z", you may consider replacing it with "+00:00" so that we retain the timezone information

svrooij:
It really helps that I'm now able to run all the tests locally in Python 3.9, only 1 error left and that is because an assertion tried to parse a datetime string with a z to compare

baywet (approved):
CI passes!! 🚀🚀🚀🚀
Thanks for the work here @svrooij and @danieleds !

PR #456, opened 2025-01-17
docs: DevContainer added and explained

### LINKED ISSUE #453: Provide DevContainer with tests configured
When trying to contribute to this project I'm faced with multiple issues (linting, testing, eg.) that I currently cannot run on my local machine.

It would be super useful if there was a devcontainer configuration and some vscode tasks to do the linting and the checking, that is now only don't in the pipeline. This would greatly help other developers that want to contribute to this project.

svrooij:
## Overview

Trying to make life easier for new contributors to this project by adding a dev container and additional docs Fixed Provide DevContainer with tests configured #453

## Related Issue

Fixes #453

### Demo

image

### Notes

Integration with the test extension in VSCode still does not work, this would be an even better addition. Has to do with poetry and the test extension, that don't play nice together.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/python
+{
+	"name": "Python 3",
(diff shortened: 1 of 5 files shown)


andrueastman (approved):
👍🏼

PR #458, opened 2025-01-26
Removed unused urllib3 dependencies

### LINKED ISSUE #452: Lower requirement for urllib
Are the dependency requirements set in stone? Or can we lower `urllib` and `httpx` to something that works with these package constrains?

https://github.com/microsoft/kiota-python/blob/dbda749947f630a03bcd5b0311fbc072182ba355/packages/http/httpx/pyproject.toml#L31-L32

urllib has a requirement of `^2.2.2` and I was wondering if we could adjust it to something that works with the following package constrain `urllib3>=1.26.5,<2`

for httpx, it seems the `http2` extras are not referenced directly, so maybe the references can be changed to `httpx>=0.28.0` and provide an extra if you want to talk to http2 servers? Like [this](https://github.com/encode/httpx/blob/c7c13f18a5af4c64c649881b2fe8dbd72a519c32/pyproject.toml#L49-L51)

Reason is, I've built this library for the Volvo Connected Vehicle api [svrooij/py_volvo_connected](https://github.com/svrooij/py_volvo_connected) that I want to use it in **Home Assistant** which has [this package contrains](https://github.com/home-assistant/core/blob/dev/homeassistant/package_constraints.txt)

Without lowering the requirement (and #448), I would not be able to use kiota for this project and would be forced to go look for something else.

svrooij:
## Overview

We were not sure how this dependency came to be read full comment [here](https://github.com/microsoft/kiota-python/issues/452#issuecomment-2602687304)

## Related Issue

Fixes #452 

### Notes

All tests still work, and the default ssl version is used from python core `ssl` to be exact. By not configuring it, it will use the default of the used python version (SSL is blocked, TLS is allowed). The `PoolManager` (which was not used), specified that it would only use tls1.2, but I don't see a reason why a library that is used to connect to all sort of apis would require something else then the platform it runs on.

> Setting the minimum TLS version is not the business of a library that is used to connect to all sort of APIs, even those that are not under control of the one connecting to them. It should be the business of the companies that provide the api.
> @svrooij => personal opinion

ps. Added poetry.lock to gitignore so people (like me) are not accidently commiting it.

## Testing Instructions

Upgrade to this version and you're good to go


baywet:
@andrueastman I'm not sure why we didn't get a release please PR for this one?

PR #460, opened 2025-01-26
Ensure 304 status code does not throw, but returns None instead

### LINKED ISSUE #363: ensure 3XX responses without location header do not throw
follow up to https://github.com/microsoft/kiota/pull/4367
and https://github.com/microsoft/kiota/issues/4190

Implement a unit test with a 304 response, no location header, check the request adapter returns null and doesn't throw.

svrooij:
Fixes #363

Add handling for 304 responses without a location header in request adapter.

* Modify `packages/http/httpx/kiota_http/httpx_request_adapter.py` to include a check for 304 status code in `_should_return_none` method.
  * Return True if the status code is 304 and there is no location header.
* Add a unit test in `packages/http/httpx/tests/test_httpx_request_adapter.py` to verify that the request adapter returns null and does not throw for a 304 response without a location header.
  * Create a mock 304 response without a location header.
  * Ensure the unit test checks that the request adapter returns null for the 304 response.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/kiota-python/pull/460?shareId=b8cffb0f-568e-4b4b-8a71-fefae2b14284).

### DIFF
--- a/packages/http/httpx/kiota_http/httpx_request_adapter.py
+++ b/packages/http/httpx/kiota_http/httpx_request_adapter.py
@@ -463,6 +463,7 @@ class HttpxRequestAdapter(RequestAdapter):
             attribute_span.record_exception(exc)
             _throw_failed_resp_span.end()
             raise exc
+        return True
 
     async def throw_failed_responses(
         self,
@@ -473,7 +474,7 @@ class HttpxRequestAdapter(RequestAdapter):
     ) -> None:
         if response.is_success or response.status_code == 304:
             return
-        if self._is_redirect_missing_location(response, parent_span, attribute_span) == False:
+        if self._is_redirect_missing_location(response, parent_span, attribute_span) is False:
             return
         try:
             attribute_span.set_status(trace.StatusCode.ERROR)
--- a/packages/http/httpx/kiota_http/httpx_request_adapter.py
+++ b/packages/http/httpx/kiota_http/httpx_request_adapter.py
@@ -440,6 +440,28 @@ class HttpxRequestAdapter(RequestAdapter):
             response.content
         ) or (not response.headers.get("location") and response.status_code in [301, 302])
 
+    def _throw_empty_redirects(self, response: httpx.Response, parent_span: trace.Span, attribute_span: trace.Span) -> None:
+        if response.is_redirect:
+            if response.has_redirect_location:
+                return
+            # Raise a more specific error if the server returned a redirect status code
+            # without a location header
+            attribute_span.set_status(trace.StatusCode.ERROR)
+            _throw_failed_resp_span = self._start_local_tracing_span(
+                "throw_failed_responses", parent_span
+            )
+            _throw_failed_resp_span.set_attribute("status", response.status_code)
+            exc = APIError(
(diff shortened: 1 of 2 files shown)


packages/http/httpx/kiota_http/httpx_request_adapter.py
```diff
@@ -423,7 +423,7 @@ async def get_root_parse_node(
             span.end()
 
     def _should_return_none(self, response: httpx.Response) -> bool:
-        return response.status_code == 204 or not bool(response.content)
+        return response.status_code == 204 or response.status_code == 304 or not bool(response.content)
```
baywet, line 426: We should also check no location header present here when the status code is 304

baywet (changes requested):

packages/http/httpx/kiota_http/httpx_request_adapter.py
svrooij: `304` means `Not Modified` according to the specs there is no redirect for this status code.

Not sure why it should validate there is no Location header.

Or so you mean if the status code is 301 or 302 and there is no Location it should return None?
In that case I would expect an error like "API wants to redirect but there is no location header"

svrooij: It is focusing on that because I focused on that.
The python library actually checks if the status code is `204` or `304` or if the response body is empty.

| StatusCode | Location header | Body | Expected |
| ---------- | --------------- | ---- | -------- |
| `204` | irrelevant | irr. | Response `None` |
| `304` | irrelevant | irr. | Response `None` |
| `301` / `302` / `303` / `307` / `308` | ✅ | irr. | Redirect |
| `301` / `302` / `303` / `307` / `308` | ❌ | ✅ | Return body? |
| `301` / `302` / `303` / `307` / `308` | ❌ | ❌ | Response `None` (or throw) |

Because if checks for the body being empty I would suspect the rest of code would either throw if it cannot redirect because of missing location. How are redirects being handled? Should this return None and then middleware takes care of the rest. In that case I completely understand the requirement for it returning None.

svrooij: in that case I can add ` or (not response.headers.get("Location") and response.status_code in [301, 302, 303, 307, 308])` if you want?

svrooij: in that case I can add  `or (not response.headers.get("Location") and response.status_code in [301, 302, 303, 307, 308])` if you want?

baywet: yes but only for 301 and 302 to align with the dotnet implementation

baywet (dismissed):

baywet (approved):
geee... finally. Python linting is quite strict.

baywet:
> > geee... finally. Python linting is quite strict.
> 
> Which is why I added the tasks to the dev container so you can run them locally 😄

yeah I was too lazy for that, run the tool, push, let the CI tell me what's wrong while I do something else. A bit noisy for others though...

PR #473, opened 2025-03-22
serialize objects to bytes in a JSON compatible way

GeorgeOG:
serialize objects to bytes in a JSON compatible way

## Overview

Before, if an API returned an object which had been deserilised, and a library serialised it to bytes afterwards, that byte string would not be valid json as python's `str` serialises dictionaries with single quotes.

## Related Issue

https://github.com/microsoftgraph/msgraph-sdk-python-core/issues/807

### Demo

Optional. Screenshots, `curl` examples, etc.

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/packages/serialization/json/kiota_serialization_json/json_parse_node.py
+++ b/packages/serialization/json/kiota_serialization_json/json_parse_node.py
@@ -233,7 +233,7 @@ class JsonParseNode(ParseNode):
         Returns:
             bytearray: The bytearray value from the nodes
         """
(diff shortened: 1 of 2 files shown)


rkodev (approved):

baywet:
@rkodev this should have been squashed with a conventional commit message to trigger the release. 
Can you create a new PR with a conventional commit please?

PR #482, opened 2025-06-23
fix: grab the first value if the server response contains a comma

MarcoP9:
## Overview

After encountering a 429 Too Many Requests Error, the server returns an incorrect Retry-After header of "30,120", resulting in an exception in the retry handler. As suggested by @musale in the comments, the function is now splitting the string by the comma and returning the first value.

## Related Issue
Contributes to fix: https://github.com/microsoftgraph/msgraph-sdk-python/issues/1159

### DIFF
--- a/packages/http/httpx/kiota_http/middleware/retry_handler.py
+++ b/packages/http/httpx/kiota_http/middleware/retry_handler.py
@@ -202,6 +202,7 @@ class RetryHandler(BaseMiddleware):
         Helper to parse Retry-After and get value in seconds.
         """
         try:
+            retry_after = retry_after.split(",")[0] if "," in retry_after else retry_after
             delay = int(retry_after)
         except ValueError:
             # Not an integer? Try HTTP date


MarcoP9:
Hi @baywet, I have added a test, I hope it is conforming to what you were expecting, let me know if I need to change something :)!

baywet:
@MarcoP9 can you have a look at the failure for the http unit tests please?

https://github.com/microsoft/kiota-python/actions/runs/15881670608/job/44846104287?pr=482

Disregard the failing serialization CI. I've authored #483 to address this.

baywet:
@MarcoP9 this looks like a narrow implementation. What of we don't exactly have the same number of digits?
Also regex are usually expensive in terms of compute.
Couldn't we simply try to parse the string as a float instead of an int?

MarcoP9:
@baywet I thought about that too, however, the server is returning a retry-after value `30,120`, containing a comma, and not a dot. A float casting would fail in that case :/. I thought about other possibilities as:
- Replacing commas with dots, but it would probably fail with http date as `Tue, 15 Nov 1994 08:12:31 GMT`
- We can split by comma and check if in the first element is all digits: `value.split(",")[0] if "," in value and value.split(",")[0].isdigit()`
What do you think?

baywet (approved):

PR #492, opened 2025-08-20
fix: ensures BaseMiddleware creates a new span instead of returning the current span

LukeMaisel:
## Overview

Previously, the _create_observability_span method in the BaseMiddleware would return the current span if no parent_span was given in the request options. Since everything that uses this method treats the returned value as a newly created span, this led to unintentionally modifying and/or ending outside spans that are not directly a part of the middleware/kiota.

I updated this method to, in the event no parent_span is given in the request, create a new span with the middleware's parent_span context or with the current context if needed. This ensures the returned span was created by the method no matter the context.

I also added another test for BaseMiddleware that ensures BaseMiddleware does not end outer spans.

## Related Issue

N/A

## Testing Instructions

* New and modified tests pass
* Other test results identical to main branch


LukeMaisel:

PR #496, opened 2025-09-01
Allow ints to be written as floats in json_serializer

### LINKED ISSUE #495: Improve int/float handling in json_serialization_writer
Setting a float field of a 'Parsable' class to an integer value is legal in python, but serializing the class then ignores th field:

Example
```python
@dataclass
class Entity(Parsable):
    # The value property
    value: Optional[float] = None

  def get_field_deserializers(self,) -> dict[str, Callable[[ParseNode], None]]:
      """
      The deserialization information for the current model
      Returns: dict[str, Callable[[ParseNode], None]]
      """
      fields: dict[str, Callable[[Any], None]] = {
          "value": lambda n : setattr(self, 'value', n.get_float_value()),
      }

   def serialize(self,writer: SerializationWriter) -> None:
        """
        Serializes information the current object
        param writer: Serialization writer to use to serialize this model
        Returns: None
        """
        if writer is None:
            raise TypeError("writer cannot be null.")
        writer.write_float_value("value", self.value) # <- int value is NOT WRITTEN 

entity = Entity(value=5) # <- int value is NOT WRITTEN 
entity = Entity(value=5.) # <- float value is WRITTEN 

```

can we just cast the passes value as float here in `json_serialization_writer.py`

```python
  def write_float_value(self, key: Optional[str], value: Optional[float]) -> None:
      """Writes the specified float value to the stream with an optional given key.
      Args:
          key (Optional[str]): The key to be used for the written value. May be null.
          value (Optional[float]): The float value to be written.
      """
      if isinstance(value, float): # type checking should catch this or we can check for int as well
          if key:
              self.writer[key] = value
          else:
              self.value = value

```

pangeorg:
## Overview

Allows ints to be set on float fields without leaving out their serialization

## Related Issue

Fixes #495

### DIFF
--- a/packages/serialization/json/tests/unit/test_json_serialization_writer.py
+++ b/packages/serialization/json/tests/unit/test_json_serialization_writer.py
@@ -11,9 +11,7 @@ from ..helpers import OfficeLocation, User, User2
 @pytest.fixture
 def user_1():
     user = User()
-    user.updated_at = datetime(2022, 1, 27, 12, 59, 45, 596117).fromisoformat(
-        "2022-01-27T12:59:45.596117+00:00"
-    )
+    user.updated_at = datetime(2022,1,27,12,59,45,596117).fromisoformat("2022-01-27T12:59:45.596117+00:00")
     user.is_active = True
     user.id = UUID("8f841f30-e6e3-439a-a812-ebd369559c36")
     return user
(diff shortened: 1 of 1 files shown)


baywet (changes requested):
Thanks for the contribution!

Can you please add unit tests to prevent future regressions?

baywet:
@pangeorg can you please sign the CLA (see the bot's message) so we can merge your pull request?

PR #516, opened 2026-03-12
fix(serialization-json): handle CSV string in get_collection_of_enum_values for flags enums

### LINKED ISSUE #515: get_collection_of_enum_values silently returns empty list for flags enums serialized as CSV string
### Description

`JsonParseNode.get_collection_of_enum_values()` in `microsoft-kiota-serialization-json` only handles `list` inputs. When the API returns a flags enum serialized as a comma-separated string (e.g. `"deviceCodeFlow"` or `"deviceCodeFlow,authenticationTransfer"`), the method silently returns an empty list `[]` instead of parsing the enum values.

This directly affects the Microsoft Graph SDK's `ConditionalAccessAuthenticationFlows.transferMethods` property, which is always deserialized as `[]` regardless of the actual API response — making it impossible to read Conditional Access policy authentication flow conditions.

### Root Cause

The `conditionalAccessTransferMethods` enum is defined in the [OpenAPI metadata](https://github.com/microsoftgraph/msgraph-metadata) with `x-ms-enum-flags: isFlags: true`:

```yaml
microsoft.graph.conditionalAccessTransferMethods:
  title: conditionalAccessTransferMethods
  enum:
    - none
    - deviceCodeFlow
    - authenticationTransfer
    - unknownFutureValue
  type: string
  x-ms-enum-flags:
    isFlags: true
```

Because of `isFlags: true`, Kiota generates `get_collection_of_enum_values()` as the deserializer (instead of `get_enum_value()`), which is correct per the design in [microsoft/kiota#3237](https://github.com/microsoft/kiota/issues/3237).

However, the Microsoft Graph API serializes this flags enum as a **plain CSV string** (e.g. `"deviceCodeFlow"`), not as a JSON array. The [official documentation](https://learn.microsoft.com/en-us/graph/api/resources/conditionalaccessauthenticationflows?view=graph-rest-1.0) confirms the JSON representation is `"transferMethods": "String"`.

The current implementation of `get_collection_of_enum_values()` does not handle this case:

```python
def get_collection_of_enum_values(self, enum_class: K) -> Optional[list[K]]:
    if isinstance(self._json_node, list):
        return list(
            map(
                lambda x: self._create_new_node(x).get_enum_value(enum_class),
                self._json_node
            )
        )
    return []  # <-- always returns [] when the node is a string
```

When `self._json_node` is `"deviceCodeFlow"` (a string, not a list), `isinstance("deviceCodeFlow", list)` is `False`, and the method returns `[]` with **no warning or error**.

### Steps to Reproduce

1. Create a Conditional Access policy in Microsoft Entra ID with **Authentication Flows** enabled and **Device code flow** selected as a transfer method:

Image

2. Fetch Conditional Access policies via the Microsoft Graph Python SDK:
   ```python
   policies = await client.identity.conditional_access.policies.get()
   for policy in policies.value:
       auth_flows = policy.conditions.authentication_flows
       if auth_flows:
           print(auth_flows.transfer_methods)  # Always prints []
   ```
3. `transfer_methods` is always `[]`, even though the policy has `deviceCodeFlow` configured

### Debug Evidence

Debug logs from a live Microsoft Graph API call (SDK v1.23.0) confirm the issue:

```
DEBUG policy 'Block Device Code Flow'
  | auth_flows=ConditionalAccessAuthenticationFlows(additional_data={}, odata_type=None, transfer_methods=[])
  | transfer_methods=[] | type=<class 'list'>
  | backing_store['transferMethods']=None
```

The SDK receives a `ConditionalAccessAuthenticationFlows` object but `transfer_methods` is always `[]` and the backing store has `transferMethods=None`.

A raw JSON request to the same endpoint confirms the API returns the value correctly:

```json
{
  "conditions": {
    "authenticationFlows": {
      "transferMethods": "deviceCodeFlow"
    }
  }
}
```

### Expected Behavior

`get_collection_of_enum_values()` should parse CSV strings (e.g. `"deviceCodeFlow"` or `"deviceCodeFlow,authenticationTransfer"`) into a list of enum values, as designed in [microsoft/kiota#3237](https://github.com/microsoft/kiota/issues/3237) where it was decided that CSV would be the default serialization format for flags enums.

### Proposed Fix

I have a PR ready that adds CSV string handling to `get_collection_of_enum_values()` in `packages/serialization/json/kiota_serialization_json/json_parse_node.py`. The fix is backward-compatible — behavior for `list` inputs is unchanged, and 5 new test cases are added with all 86 tests passing.

### Environment

- `microsoft-kiota-serialization-json`: 1.9.2 (also verified on `main` / 1.9.9)
- `msgraph-sdk`: 1.23.0 (also verified against `main` / 1.55.0)
- Python: 3.12+
- Verified against live Microsoft Graph API (v1.0)

### Related

- [microsoft/kiota#3237](https://github.com/microsoft/kiota/issues/3237) — Generation/Handling of bitwise/flagged enums (design decision for CSV format)
- [microsoftgraph/msgraph-sdk-python#480](https://github.com/microsoftgraph/msgraph-sdk-python/issues/480) — Related enum deserialization issue in Conditional Access policies (different field, same root cause pattern)

HugoPBrito:
## Summary

Fixes #515

`get_collection_of_enum_values()` in `JsonParseNode` silently returns `[]` when the JSON node is a string instead of a list. This breaks deserialization of flags enums that are serialized as comma-separated strings by the API (e.g. `"deviceCodeFlow"` or `"deviceCodeFlow,authenticationTransfer"`).

## Problem

Flags enums marked with `x-ms-enum-flags: isFlags: true` in OpenAPI metadata are generated by Kiota to use `get_collection_of_enum_values()` as their deserializer ([microsoft/kiota#3237](https://github.com/microsoft/kiota/issues/3237)). However, the Microsoft Graph API serializes these as **CSV strings**, not JSON arrays.

The current code:

```python
def get_collection_of_enum_values(self, enum_class):
    if isinstance(self._json_node, list):
        return list(map(...))
    return []  # <-- string input always hits this path
```

This directly affects `ConditionalAccessAuthenticationFlows.transferMethods` in the Microsoft Graph SDK, which always deserializes as `[]` regardless of the actual API response.

## Fix

Added CSV string handling after the existing `list` check:

```python
if isinstance(self._json_node, str) and self._json_node:
    return [
        result
        for value in self._json_node.split(",")
        if (result := self._create_new_node(value.strip()).get_enum_value(enum_class)) is not None
    ]
```

The fix is **backward-compatible** — behavior for `list` inputs is completely unchanged.

## Tests

5 new test cases added covering:
- CSV string with multiple values (`"dunhill,oval"`)
- CSV string with spaces (`"dunhill, oval"`)
- Single string value (`"dunhill"`)
- Empty string (`""`)
- CSV string with unknown members (`"dunhill,unknownValue,oval"` — unknown values silently skipped)

All **86 tests pass** (81 existing + 5 new), zero regressions.

## Verification

- Verified the bug against a **live Microsoft Graph API** (v1.0) with a real Conditional Access policy that has Device Code Flow configured
- Confirmed the API returns `"transferMethods": "deviceCodeFlow"` as a string (not an array)
- Confirmed the fix resolves the issue for `ConditionalAccessAuthenticationFlows` in `msgraph-sdk` v1.23.0
- Verified the same code is still on `main` (v1.9.9)

### DIFF
--- a/packages/serialization/json/kiota_serialization_json/json_parse_node.py
+++ b/packages/serialization/json/kiota_serialization_json/json_parse_node.py
@@ -197,9 +197,9 @@ class JsonParseNode(ParseNode):
         # how flags enums are represented in OpenAPI via x-ms-enum-flags.
         if isinstance(self._json_node, str) and self._json_node:
             return [
-                result
-                for value in self._json_node.split(",")
-                if (result := self._create_new_node(value.strip()).get_enum_value(enum_class)) is not None
+                result for value in self._json_node.split(",")
+                if (result := self._create_new_node(value.strip()).get_enum_value(enum_class)
+                    ) is not None
             ]
         return []
 


baywet:
@HugoPBrito can you fix the formatting before we merge this PR please?

PR #560, opened 2026-05-28
fix(json): remove redundant guard in get_child_node

### LINKED ISSUE #559: [Bug] JsonParseNode.get_child_node raises ValueError for empty or None input
An early guard in `get_child_node` raises a `ValueError` if an empty or None input is given.

Given that the method returns an `Optional[ParseNode]`, this is unnecessary, since it can just return `None` without breaking any documented behavior.
Besides this, it breaks auto-generated code which calls `get_child_node("")` when there is no discriminator which is why I marked it as Bug.

The proposed solution is to simply remove that guard.

gntikos:
## Overview

Removes an identified guard in `get_child_node`. The guard does not really protect from anything, but breaks generated code, for example in cases of `oneOf` without discriminator (see [`microsoft/kiota#6868`](https://github.com/microsoft/kiota/issues/6868)).

The method now returns `None`, which is allowed since the return type is `Optional[ParseNode]`.

## Related Issue

Fixes #559

### DIFF
--- a/packages/serialization/json/tests/unit/test_json_parse_node.py
+++ b/packages/serialization/json/tests/unit/test_json_parse_node.py
@@ -35,6 +35,19 @@ def test_get_float_value_from_float():
     assert result == 44.6
 
 
+def test_get_child_node_not_found():
+    parse_node = JsonParseNode({"name": "Diego Siciliani"})
+    result = parse_node.get_child_node("age")
+    assert result is None
+
+
+def test_get_child_node_empty_name():
+    parse_node = JsonParseNode({"": "Diego Siciliani"})
+    result = parse_node.get_child_node("")
+    assert result is not None
+    assert result.get_str_value() == "Diego Siciliani"
+
+
 @pytest.mark.parametrize("value", [0, 10, 100])
 def test_get_float_value(value: int):
     """
--- a/packages/serialization/json/tests/unit/test_json_parse_node.py
+++ b/packages/serialization/json/tests/unit/test_json_parse_node.py
@@ -36,16 +36,16 @@ def test_get_float_value_from_float():
 
 
 def test_get_child_node_not_found():
-    parse_node = JsonParseNode({"name": "Diego Siciliani"})
+    parse_node = JsonParseNode({"name": "Jane Smith"})
     result = parse_node.get_child_node("age")
     assert result is None
 
 
 def test_get_child_node_empty_name():
-    parse_node = JsonParseNode({"": "Diego Siciliani"})
+    parse_node = JsonParseNode({"": "John Smith"})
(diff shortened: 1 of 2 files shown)


gntikos:
@microsoft-github-policy-service agree company="i4energy"

packages/serialization/json/tests/unit/test_json_parse_node.py
```diff
@@ -35,6 +35,19 @@ def test_get_float_value_from_float():
     assert result == 44.6
 
 
+def test_get_child_node_not_found():
+    parse_node = JsonParseNode({"name": "Diego Siciliani"})
```
baywet:
```suggestion
    parse_node = JsonParseNode({"name": "Jane Smith"})
```

packages/serialization/json/tests/unit/test_json_parse_node.py
```diff
@@ -35,6 +35,19 @@ def test_get_float_value_from_float():
...
+    result = parse_node.get_child_node("age")
+    assert result is None
+
+
+def test_get_child_node_empty_name():
+    parse_node = JsonParseNode({"": "Diego Siciliani"})
```
baywet:
```suggestion
    parse_node = JsonParseNode({"": "John Smith"})
```

packages/serialization/json/tests/unit/test_json_parse_node.py
```diff
@@ -35,6 +35,19 @@ def test_get_float_value_from_float():
...
+    result = parse_node.get_child_node("")
+    assert result is not None
+    assert result.get_str_value() == "Diego Siciliani"
```
baywet:
```suggestion
    assert result.get_str_value() == "John Smith"
```

Not sure whether it was a fictional name or not, but let's make it clear it's fictional and not related to any real person :)

baywet (changes requested):
Hi @gntikos 
Thanks for the contribution!

A few minor remarks and we should be good to go.

Let me know if you have any additional comments or questions.

packages/serialization/json/tests/unit/test_json_parse_node.py
gntikos: Sure, no problem. I just saw it in the previous test and just repeated it.

packages/serialization/json/tests/unit/test_json_parse_node.py
gntikos: done.

packages/serialization/json/tests/unit/test_json_parse_node.py
gntikos: done

baywet (approved):

PR #563, opened 2026-06-04
chore: drop poetry, migrate to pip + flit_core

MIchaelMainer:
It was highly suggested from an internal Pythonista that we should drop Poetry as it adds extra complexity when encountering edge scenarios. This will also make Kiota-Python better match what we are doing for the graph repos.

- Replace poetry-core build backend with flit_core in all 8 packages
- Convert all pyproject.toml from [tool.poetry] to PEP 621 [project] format
- Fix [mypy] -> [tool.mypy] in abstractions, authentication/azure, bundle, tests/validation
- Fix tests/validation mypy files setting to 'validation'
- Create requirements-dev.txt per package (8 packages + tests/validation)
- Update build.yml: replace poetry install/run with pip and direct commands
- Update cd-publish-python.yml: replace poetry with pip install build + python -m build
- Remove poetry-plugin-mono-repo-deps from root requirements_dev.txt

Validation runs:
https://github.com/microsoft/kiota-python/actions/runs/26969442452
https://dev.azure.com/microsoftgraph/Graph%20Developer%20Experiences/_build/results?buildId=221616&view=results


packages/bundle/requirements-dev.txt
```diff
@@ -0,0 +1,13 @@
+-e ../abstractions
+-e ../http/httpx
+-e ../serialization/json
+-e ../serialization/form
+-e ../serialization/text
+-e ../serialization/multipart
```
baywet, line 6: one thing that we should validate before merging this is whether a "publish" (to local file) leads to the package metadata being emitted with the corresponding dependency, instead of local reference. This would otherwise lead to broken packages and has bitten us a couple of times in the past.

And I don't know enough about Python to "tell that from just looking at the configuration"

MIchaelMainer: I don't think this will be a problem as those are dev dependencies. The dependencies in the project file point to the packages

```
dependencies = [
    "microsoft-kiota-abstractions>=1.10.2,<2.0.0",
    "microsoft-kiota-http>=1.10.2,<2.0.0",
    "microsoft-kiota-serialization-json>=1.10.2,<2.0.0",
    "microsoft-kiota-serialization-form>=1.10.2,<2.0.0",
    "microsoft-kiota-serialization-text>=1.10.2,<2.0.0",
    "microsoft-kiota-serialization-multipart>=1.10.2,<2.0.0",
]
```

PR #565, opened 2026-06-05
fix: close streaming responses before retrying

### LINKED ISSUE #564: [Bug] Retry handling does not close retried responses, which leaks httpx connections
**Environment**

- Python Version: 3.12.12
- kiota-http version: 1.10.1
- OS: Debian GNU/Linux 12 (bookworm), running in Kubernetes
- Consumed via: msgraph-sdk 1.58.0 (msgraph-core 1.4.0)
- HTTP stack: httpx 0.27.2 / httpcore 1.0.9

Seen during long, retry-heavy Microsoft Graph ingestions via `msgraph-sdk-python`.

**Describe the bug**

[RetryHandler.send](https://github.com/microsoft/kiota-python/blob/main/packages/http/httpx/kiota_http/middleware/retry_handler.py#L69-L109) retries on 429/503/504, but it reassigns `response` to the
next attempt's response without closing the previous one. Because the
handler runs at the transport/middleware layer (below `AsyncClient.send`'s
auto-read/close of the final response), each retried response is a streamed,
unread response whose connection is never returned to the pool. Over a
long-running, retry-heavy workload the connection pool is exhausted and
subsequent requests fail with `httpx.PoolTimeout`.

This is the same class of bug that was fixed for `RedirectHandler` in
microsoft/kiota-http-python#299 (PR microsoft/kiota-http-python#300), where the
fix was an `await response.aclose()` before discarding the old response.

**To Reproduce**

Self-contained, end-to-end repro against a real connection pool: a local server
returns a retryable `429`, the default kiota middleware retries, and a wrapping
transport records the responses the pool hands back. The retried responses are
never closed (`is_closed == False`), so their connections are never returned to
the pool, and once it is exhausted the request raises `httpx.PoolTimeout`:

```python
import asyncio
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

import httpx
from kiota_http.kiota_client_factory import KiotaClientFactory

PORT = 8807


class QuietServer(ThreadingHTTPServer):
    def handle_error(self, request, client_address):
        pass  # silence benign ConnectionReset when the client drops the socket


class Handler(BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"  # keep-alive, so the pool tracks the connection

    def log_message(self, *a):
        pass

    def do_GET(self):
        body = b"{}"
        self.send_response(429)  # retryable -> kiota RetryHandler retries
        self.send_header("Retry-After", "0")
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)


class RecordingTransport(httpx.AsyncBaseTransport):
    """Captures every response from the real pool so we can inspect it after."""

    def __init__(self, inner):
        self.inner = inner
        self.responses = []

    async def handle_async_request(self, request):
        response = await self.inner.handle_async_request(request)
        self.responses.append(response)
        return response

    async def aclose(self):
        await self.inner.aclose()


async def main():
    threading.Thread(
        target=QuietServer(("127.0.0.1", PORT), Handler).serve_forever,
        daemon=True,
    ).start()

    # real pool of 2; record the responses the pool hands back
    recorder = RecordingTransport(
        httpx.AsyncHTTPTransport(
            limits=httpx.Limits(max_connections=2, max_keepalive_connections=2)
        )
    )
    client = httpx.AsyncClient(transport=recorder, timeout=httpx.Timeout(5.0, pool=2.0))
    client = KiotaClientFactory.create_with_default_middleware(client=client)

    try:
        await client.get(f"http://127.0.0.1:{PORT}/")
        print("completed without error (unexpected)")
    except httpx.PoolTimeout:
        print("retried responses is_closed:", [r.is_closed for r in recorder.responses])
        print("httpx.PoolTimeout raised once the pool is exhausted by the leaked "
              "(unclosed) connections")
    finally:
        await client.aclose()


asyncio.run(main())
```
Output:

```
retried responses is_closed: [False, False]
httpx.PoolTimeout raised once the pool is exhausted by the leaked (unclosed) connections
```

The leaked connections sit checked-out but idle (no `ReadTimeout`). In
production this is gradual: with a larger pool each retried response leaks one
connection until the pool is exhausted.

**Proposed Fix**
Add an `await response.aclose()` after line 97 in the retry handler. Created this issue per the CONTRIBUTING.md and will follow up with a PR referencing this issue.

Kashyap456:
## Overview

This PR closes a response in RetryHandler's retry loop before response gets reassigned by the retry loop. This allows the connection that produced the response to get released back to its connection pool to serve another request.

## Related Issue

Fixes #564

## Testing Instructions

Added test_retried_responses_are_closed to packages/http/httpx/tests/middleware_tests/test_retry_handler.py to verify retried responses are closed (so their connections are released back to the pool). It drives the real RetryHandler with a MockTransport returning 429, 429, 200, and asserts the two retried responses report is_closed while the returned 200 stays open. Runs in the existing suite via pytest.

### DIFF
--- a/packages/http/httpx/kiota_http/middleware/retry_handler.py
+++ b/packages/http/httpx/kiota_http/middleware/retry_handler.py
@@ -95,6 +95,8 @@ class RetryHandler(BaseMiddleware):
             # and status code
             should_retry = self.should_retry(request, current_options, response)
             if all([should_retry, retry_valid, delay < RetryHandlerOption.MAX_DELAY]):
+                # release the discarded response's connection before retrying
+                await response.aclose()
                 await asyncio.sleep(delay)
                 # increment the count for retries
                 retry_count += 1
--- a/packages/http/httpx/tests/middleware_tests/test_retry_handler.py
+++ b/packages/http/httpx/tests/middleware_tests/test_retry_handler.py
@@ -298,3 +298,32 @@ async def test_retry_options_apply_per_request():
     assert resp.status_code == GATEWAY_TIMEOUT
     assert 'request_2' in resp.request.headers
     assert resp.request.headers[RETRY_ATTEMPT] == '2'
+
+
+@pytest.mark.asyncio
+async def test_retried_responses_are_closed():
+    """Discarded retry responses are closed so their connection is released."""
+    responses = []
+
+    async def body():
+        yield b'{}'
+
+    statuses = [TOO_MANY_REQUESTS, TOO_MANY_REQUESTS, 200]
+
+    def request_handler(request: httpx.Request):
+        status = statuses.pop(0)
+        headers = {RETRY_AFTER: '0'} if status == TOO_MANY_REQUESTS else {}
+        response = httpx.Response(status, headers=headers, content=body())
+        responses.append(response)
+        return response
+
+    handler = RetryHandler(RetryHandlerOption(delay=0, max_retries=2))
+    request = httpx.Request('GET', BASE_URL)
+    mock_transport = httpx.MockTransport(request_handler)
+    resp = await handler.send(request, mock_transport)
+
+    assert resp.status_code == 200
+    # the two retried responses are closed, releasing their connections
+    assert responses[0].is_closed and responses[1].is_closed
+    # the returned response is left open for the caller to read
+    assert not responses[2].is_closed


baywet (approved):
Thanks for the contribution! Please make sure you sign the CLA (see the bot message)

Kashyap456:
sorry for the delay here, just making sure I have the right approvals to sign the CLA on behalf of my company, because I made this change as part of work

baywet:
No worries, we're working on a deployment pipeline migration anyway.

Kashyap456:

Kashyap456:
I was advised by my employer to sign on my own behalf.

PR #573, opened 2026-06-16
fix: validate URL host against allowed hosts whitelist

ramsessanchez:
## Summary

Fixes a bug in \AllowedHostsValidator.is_url_host_valid()\ where the method was **not actually validating** the URL's host against the configured allowed hosts.

## Problem

The method only checked that the URL had a valid scheme and netloc:

\\\python
o = urlparse(url)
return all([o.scheme, o.netloc])
\\\

This meant **any** well-formed URL would pass validation, completely bypassing the host whitelist. Authentication tokens could be sent to arbitrary hosts.

## Fix

The method now extracts the hostname and checks it against the \llowed_hosts\ set:

\\\python
o = urlparse(url)
if not all([o.scheme, o.netloc]):
    return False
return o.hostname.lower() in self.allowed_hosts
\\\

## Tests Added

Added comprehensive unit tests covering:
- Allowed vs disallowed hosts
- Case-insensitive matching
- Empty allowed hosts list (all hosts valid)
- URLs with ports, query params, fragments
- Subdomain validation (strict match)
- Invalid/missing scheme or netloc
- \set_allowed_hosts\ behavior


ramsessanchez:
Ignore quality gate warning here: error is complaining about http vs https in a validation test, however, http is used as it is supposed to raise an error for a malformed host. This case should fail host validation.

MIchaelMainer (approved):

MIchaelMainer (approved):

PR #646, opened 2026-06-29
fix(json): base64-decode bytes values on deserialization

### LINKED ISSUE #636: JsonParseNode.get_bytes_value() does not base64-decode Edm.Binary (asymmetric with write_bytes_value)
### Description

`JsonParseNode.get_bytes_value()` does not base64-decode `Edm.Binary` values. It returns the **base64 string encoded to UTF-8**, i.e. the raw base64 text as bytes, instead of the decoded binary.

This is asymmetric with `JsonSerializationWriter.write_bytes_value()`, which **does** base64-encode. So a `bytes` value written by kiota cannot be round-tripped back through kiota — the read side gives you the base64 text, not the original bytes.

### Versions

- `microsoft-kiota-serialization-json` 1.10.3 (behavior is identical on `main`)

### Current implementation

Reader — `packages/serialization/json/kiota_serialization_json/json_parse_node.py`:

```python
@staticmethod
def _get_bytes_value(value: Any) -> Optional[bytes]:
    # if the node is a string, we need to decode it
    # This ensures that the string is properly converted to bytes
    if isinstance(value, str):
        base64_string = value
    else:
        base64_string = json.dumps(value)
    if not base64_string:
        return None
    return base64_string.encode("utf-8")   # <-- never base64-decodes
```

The comment says "we need to decode it", but the code only UTF-8-encodes the base64 string.

Writer — `packages/serialization/json/kiota_serialization_json/json_serialization_writer.py` (correct, for contrast):

```python
def write_bytes_value(self, key, value):
    if isinstance(value, bytes):
        base64_string = base64.b64encode(value).decode("utf-8")
        ...
```

### Reproduction

```python
import base64
from kiota_serialization_json.json_serialization_writer import JsonSerializationWriter
from kiota_serialization_json.json_parse_node import JsonParseNode

raw = b"PK\x03\x04hello"  # arbitrary binary

w = JsonSerializationWriter()
w.write_bytes_value("contentBytes", raw)
serialized = w.writer["contentBytes"]      # 'UEsDBGhlbGxv'  (base64 text — correct)

got = JsonParseNode(serialized).get_bytes_value()
print(repr(got))                           # b'UEsDBGhlbGxv'  <-- base64 TEXT, not the bytes
print(got == raw)                          # False
print(got == serialized.encode("utf-8"))   # True
```

**Expected:** `get_bytes_value()` returns `b"PK\x03\x04hello"`.
**Actual:** it returns `b"UEsDBGhlbGxv"` (the base64 string as UTF-8 bytes).

### Impact

Any `Edm.Binary` property deserialized via this code path is wrong. A concrete real-world case: Microsoft Graph `fileAttachment.contentBytes` comes back as the base64 text rather than the file. Saving `attachment.content_bytes` to disk yields a file that is the base64 of the real file — e.g. a downloaded `.xlsx` fails to open because the bytes are ASCII base64 (`UEsDBBQA...`), not the zip (`PK\x03\x04...`).

### Suggested fix

Base64-decode in `_get_bytes_value`, mirroring the writer:

```python
@staticmethod
def _get_bytes_value(value: Any) -> Optional[bytes]:
    if not isinstance(value, str) or not value:
        return None
    return base64.b64decode(value)
```

Zelys-DFKH:
## Overview

`JsonParseNode.get_bytes_value()` was returning the raw base64 string re-encoded as UTF-8 bytes instead of decoding it. So any `Edm.Binary` property from Microsoft Graph (like `fileAttachment.contentBytes`) would come back as the base64 text in bytes form rather than the actual file data. The serialization writer already uses `base64.b64encode` correctly, so the reader wasn't matching it.

The fix replaces the broken body with `base64.b64decode(value)` for string inputs and `None` for non-strings. This mirrors how the text-format parse node handles the same field.

## Related Issue

Fixes #636

## Testing Instructions

- `cd packages/serialization/json`
- `pip install -e ".[dev]"` (or `poetry install`) if not already set up
- Run the bytes-specific tests: `pytest tests/unit/test_json_parse_node.py -k bytes -v`
- All three tests should pass: `test_get_bytes_value`, `test_get_bytes_value_empty_string`, and `test_get_bytes_value_non_string_returns_none`
- Full suite: `pytest` (89 tests, all pass)

### Notes

Thanks to @RockyMM for the detailed bug report, precise code location, and suggested fix in #636. This PR implements exactly that approach.

### DIFF
--- a/packages/serialization/json/kiota_serialization_json/json_parse_node.py
+++ b/packages/serialization/json/kiota_serialization_json/json_parse_node.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import json
 import re
 import warnings
@@ -290,15 +291,9 @@ class JsonParseNode(ParseNode):
 
     @staticmethod
     def _get_bytes_value(value: Any) -> Optional[bytes]:
-        # if the node is a string, we need to decode it
-        # This ensures that the string is properly converted to bytes
         if isinstance(value, str):
-            base64_string = value
-        else:
-            base64_string = json.dumps(value)
-        if not base64_string:
-            return None
-        return base64_string.encode("utf-8")
+            return base64.b64decode(value)
+        return None
 
     @property
     def on_before_assign_field_values(self) -> Optional[Callable[[Parsable], None]]:
--- a/packages/serialization/json/tests/unit/test_json_parse_node.py
+++ b/packages/serialization/json/tests/unit/test_json_parse_node.py
@@ -1,3 +1,4 @@
+import base64
 import json
 from datetime import date, datetime, time, timedelta, timezone
 from uuid import UUID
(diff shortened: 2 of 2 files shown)


baywet (commented):
Thanks for the contribution!

Would you mind addressing copilot's comment please? (If I do it, I won't be able to approve/merge). Besides that, we should be good to go.

baywet (approved):

PR #664, opened 2026-07-18
fix(json): only convert unambiguous date/datetime/UUID strings in additional data

### LINKED ISSUE #487: try_get_anything - converts a 32 hex string to UUID
The library is modifying the content of payload just because the content has something that can taken by UUID class.

This is failing the schema validation down the road. 

The location of the culprit

https://github.com/microsoft/kiota-python/blob/8740fcb17ddc2e943baaecfaba98bfef97ef920a/packages/serialization/json/kiota_serialization_json/json_parse_node.py#L331

56badkarma:
## Overview

`JsonParseNode.try_get_anything` runs every untyped string (everything landing in `additional_data`) through five lenient parsers in sequence: `datetime_from_iso_format_compat`, `UUID`, `parse_timedelta_string`, `date.fromisoformat`, `time_from_iso_format_compat`. On Python 3.11+ the lenient parsers accept many ordinary text values:

```
'19-2026'       -> datetime.time(19, 0, tzinfo=timezone(timedelta(days=-1, seconds=12840)))  # invoice number
'100-012863299' -> datetime.time(10, 0, ...)                                                 # invoice number
'PT'            -> datetime.timedelta(0)                                                     # country code
'10:30'         -> datetime.timedelta(seconds=37800)
'11.0'          -> datetime.time(11, 0)                                                      # version string
'd41d8cd98f00b204e9800998ecf8427e' -> UUID(...)                                              # 32-hex string
```

This is not just a display problem: in our production SharePoint AP-invoicing app, invoice numbers like `19-2026` were coerced to `time` objects, later stringified on update, and persisted back — permanently corrupting stored data (details with a captured wire payload in microsoftgraph/msgraph-sdk-python#1340).

This PR replaces the speculative parsing with two anchored patterns, mirroring the .NET [`JsonParseNode.TryGetAnything`](https://github.com/microsoft/kiota-dotnet/blob/main/src/serialization/json/JsonParseNode.cs#L510), which only attempts strict `DateTime`/`DateTimeOffset`/`Guid` parses and otherwise returns the string:

- full ISO date or datetime (`^\d{4}-\d{2}-\d{2}([Tt ]\d{2}:\d{2}|$)`) → `datetime_from_iso_format_compat`
- canonical hyphenated UUID → `UUID`

Behavior is unchanged for real payload values: full ISO datetimes still become `datetime`, date-only strings still become `datetime` (same as today, via the same compat parser), canonical UUIDs still become `UUID`. The `timedelta`/`time` guessing is removed entirely — Graph does not send bare durations or times as untyped strings, and their grammars overlap with too much real-world text. The `__is_four_digit_number`/`isdigit()` special cases (added for earlier symptom reports like microsoftgraph/msgraph-sdk-python#653) become unnecessary since all-digit strings cannot match either anchored pattern, and are removed.

## Related Issue

Fixes #487
Addresses microsoftgraph/msgraph-sdk-python#1340 and microsoftgraph/msgraph-sdk-python#653

## Notes

- Strings that previously coerced to `timedelta`/`time`/unhyphenated-UUID will now stay strings in `additional_data`. That is the intent of the fix, but callers who relied on those accidental conversions would see a change; typed model fields are unaffected (they go through declared field deserializers and never reach `try_get_anything`).
- Verified against a live Microsoft Graph SharePoint tenant: text columns round-trip unchanged, `Created`/`Modified` datetimes still parse.

## Testing Instructions

* `cd packages/serialization/json && pytest tests/`
* New parametrized test `test_get_anything_does_not_coerce_text_values` covers each reported corruption pattern (`19-2026`, `PT`, `10:30`, `23.085`, `11.0`, 32-hex, date fragments)
* New tests pin the preserved conversions: date-only → `datetime`, canonical UUID → `UUID`; existing tests cover full ISO datetimes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019Uui4g7E7UmzKRGHMW5zbH

### DIFF
--- a/packages/serialization/json/tests/unit/test_json_parse_node.py
+++ b/packages/serialization/json/tests/unit/test_json_parse_node.py
@@ -228,6 +228,47 @@ def test_get_anythin_does_convert_date_string_to_datetime():
     assert result == datetime(2023, 10, 5, 14, 48, tzinfo=timezone.utc)
 
 
+@pytest.mark.parametrize(
+    "value",
+    [
+        # microsoftgraph/msgraph-sdk-python#1340: text values coerced to
+        # datetime.time via lenient time.fromisoformat (3.11+ accepts an
+        # hour plus a UTC-offset-looking suffix or a fraction separator)
+        "19-2026",  # invoice number, parsed as time(19, 0) at offset -20:26
+        "100-012863299",  # invoice number
+        "23.085",  # parsed as time(23, 0, 0, 85000)
+        "11.0",  # version string, parsed as time(11, 0)
+        # coerced to timedelta via parse_timedelta_string
+        "PT",  # country code, parsed as an empty ISO 8601 duration
+        "10:30",  # parsed as hh:mm
+        # microsoft/kiota-python#487: 32-hex string coerced to UUID
+        "d41d8cd98f00b204e9800998ecf8427e",
+        # date-like fragments must not be parsed as dates
+        "12-25",
+        "2023-10",
+    ],
+)
+def test_get_anything_does_not_coerce_text_values(value):
+    parse_node = JsonParseNode(value)
+    result = parse_node.try_get_anything(value)
+    assert isinstance(result, str)
+    assert result == value
+
+
+def test_get_anything_converts_date_only_string_to_datetime():
+    parse_node = JsonParseNode("2023-10-05")
+    result = parse_node.try_get_anything("2023-10-05")
+    assert isinstance(result, datetime)
+    assert result == datetime(2023, 10, 5, 0, 0)
+
+
+def test_get_anything_converts_canonical_uuid_string():
+    parse_node = JsonParseNode("8f841f30-e6e3-439a-a812-ebd369559c36")
+    result = parse_node.try_get_anything("8f841f30-e6e3-439a-a812-ebd369559c36")
+    assert isinstance(result, UUID)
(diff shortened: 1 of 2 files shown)


56badkarma:

packages/serialization/json/tests/unit/test_json_parse_node.py
```diff
@@ -228,6 +228,47 @@ def test_get_anythin_does_convert_date_string_to_datetime():
...
+)
+def test_get_anything_does_not_coerce_text_values(value):
+    parse_node = JsonParseNode(value)
+    result = parse_node.try_get_anything(value)
+    assert isinstance(result, str)
+    assert result == value
+
+
+def test_get_anything_converts_date_only_string_to_datetime():
+    parse_node = JsonParseNode("2023-10-05")
```
baywet, line 259: Can we have variations with time, and time + offset to prevent any further regression?

baywet (changes requested):
Thanks for the contribution!

packages/serialization/json/tests/unit/test_json_parse_node.py
56badkarma: Added parametrized variations covering date+time with no offset, Z, +00:00, non-UTC offsets (+02:00, -05:00 with fractional seconds), lowercase t, and the RFC 3339 space separator — all pinned to still convert to datetime.

baywet (approved):

PR #666, opened 2026-07-23
Handle map-style query parameters with None values

bkoelman:
Contributes to https://github.com/microsoft/kiota/issues/3800.

Dict values used as map/dictionary query parameters (from an OpenAPI `type: object` with `additionalProperties` field) are now sanitized before being passed to StdUriTemplate.  None values are silently dropped per RFC 6570 §2.3 "undefined" semantics, preventing a ValueError when StdUriTemplate tries to convert None to a string.  An empty string must be set explicitly to send `key=`.

Adds seven test cases covering map expansion, None-value omission, empty-string inclusion, empty dicts, all-None dicts, mixed scalar+map parameters, and the set_query_string_parameters_from_raw_object() path.

### DIFF
--- a/packages/abstractions/kiota_abstractions/date_utils.py
+++ b/packages/abstractions/kiota_abstractions/date_utils.py
@@ -1,6 +1,6 @@
-from sys import version_info as sys_version_info
 import re
 from datetime import datetime, time, timedelta
+from sys import version_info as sys_version_info
 
 _ISO8601_DURATION_PATTERN = re.compile(
     "^P"  # Duration P indicator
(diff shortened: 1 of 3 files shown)


baywet (dismissed):
Thanks for the contribution!

baywet:
@bkoelman can you fix the formatting as [reported by this action run](https://github.com/microsoft/kiota-python/actions/runs/29968417013/job/89222799109?pr=666) please? after that I think we'd be good to go

bkoelman:
pushed fixes and rebased on latest main

baywet (approved):

PR #688, opened 2026-08-13
Fix/multipart stray linebreak

### LINKED ISSUE #687: [Bug] MultipartBody appends trailing newline to string part values
`MultipartBody.serialize()` writes every string part's value with `write_str_value()`,
which always appends a trailing newline. That newline therefore becomes part of the value
itself. A part meant to hold the value `"helloworld"` is sent as `"helloworld\r\n"`.
Consumers that compare that value exactly would therefore reject it. (e.g. if that value is enum-validated)

Problem:
 `MultipartBody.serialize()` (line 125) calls `writer.write_str_value("", part_value[1])` . `write_str_value()` always appends `"\n"` at the end of every call, regardless of whether a key was given.

Suggested Fix:
Replace `writer.write_str_value("", part_value[1])` with `writer.write_bytes_value("", part_value[1].encode('utf-8'))`, which preserves raw bytes but avoids inserting newline.

GitGud44:
## Overview

`MultipartBody.serialize()` writes every string part's value with `write_str_value()`,
which always appends a trailing newline. That newline therefore becomes part of the value
Itself. E.g.  A part meant to hold the value `"helloworld"` is sent as `"helloworld\r\n"`.
Consumers that compare that value exactly would then reject it.

Fixed by writing string part values with `write_bytes_value()` instead, which writes raw bytes with no terminator.

## Related Issue

Fixes #687

### Demo

Before (current `main`), a `text/plain` part with value `"Hello world"`:
```
--boundary\r\nContent-Type: text/plain\r\nContent-Disposition: form-data; name="message"\r\n\r\nHello world\r\n\r\n--boundary--\r\n
```

After (this PR):
```
--boundary\r\nContent-Type: text/plain\r\nContent-Disposition: form-data; name="message"\r\n\r\nHello world\r\n--boundary--\r\n
```

### Notes

- Only string part values are affected. `bytes`/file parts already go through
  `write_bytes_value()` and never carried the stray CRLF.
- No existing test covered a plain `str` part before this PR, every
  `test_write_object_value*` case used a JSON (`Parsable`) part and a `bytes` part, never
  a bare string one.

## Testing Instructions

* `cd packages/serialization/multipart && poetry install`
* `poetry run pytest`
** Expect `test_write_object_value_with_string_part` and
  `test_write_object_value_with_multiple_string_parts` to pass (both fail on `main`
  without this fix (double CRLF before the boundary instead of one))

### DIFF
--- a/packages/abstractions/kiota_abstractions/multipart_body.py
+++ b/packages/abstractions/kiota_abstractions/multipart_body.py
@@ -122,7 +122,7 @@ class MultipartBody(Parsable, Generic[T]):
             if isinstance(part_value[1], Parsable):
                 self._write_parsable(writer, part_value)
             elif isinstance(part_value[1], str):
-                writer.write_str_value("", part_value[1])
+                writer.write_bytes_value("", part_value[1].encode('utf-8'))
             elif isinstance(part_value[1], bytes):
                 writer.write_bytes_value("", part_value[1])
             elif isinstance(part_value[1], io.IOBase):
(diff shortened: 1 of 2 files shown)


baywet (approved):
Thanks for the contribution!

PR #701, opened 2026-08-26
Add wildcard host matching to AllowedHostsValidator

### LINKED ISSUE #700: Add support for suffix-based (wildcard) matching in AllowedHostsValidator
**Is your feature request related to a problem? Please describe.**

The `AllowedHostsValidator` currently only supports **exact host matching**. This becomes a significant limitation when working with services that use **dynamic, tenant-specific subdomains** — most notably Microsoft Fabric.

For example, Microsoft Fabric GraphQL endpoints follow this pattern:

```
https://<workspace-guid>.z2c.graphql.fabric.microsoft.com/v1/workspaces/<workspace-id>/graphqlapis/<api-id>/graphql
```

Where `<workspace-guid>` is unique per workspace (e.g. `2c662c476e2f45e0b842bb63307de046`). Since the subdomain changes per workspace, it is impossible to pre-configure the full hostname in `allowed_hosts`. In practice, this forces developers to:

1. Enumerate and register **every** workspace-specific hostname ahead of time, or
2. Set `allowed_hosts` to an empty list, which **disables host validation entirely** and silently sends bearer tokens to any host — a security risk.

Neither option is practical or safe when consuming multiple Fabric workspaces.

**Describe the solution you'd like**

Support **suffix-based (wildcard) matching** in `AllowedHostsValidator`. A host entry starting with `.` would match any subdomain under that domain. For example:

```python
AllowedHostsValidator(allowed_hosts=[".fabric.microsoft.com"])
```

Would match:
- `2c662c47.z2c.graphql.fabric.microsoft.com` ✅
- `a1b2c3d4.z2c.graphql.fabric.microsoft.com` ✅
- `fabric.microsoft.com` (the bare domain itself) ❌ — suffix matching only applies to subdomains

The implementation in `is_url_host_valid` would add a single check alongside the existing exact-match logic:

```python
def is_url_host_valid(self, url: str) -> bool:
    if not url:
        return False
    if not self.get_allowed_hosts():
        return True
    o = urlparse(url)
    if not all([o.scheme, o.netloc]):
        return False
    if not o.hostname:
        return False
    hostname = o.hostname.lower()
    # Existing exact match
    if hostname in self.allowed_hosts:
        return True
    # Suffix match for entries starting with "."
    return any(
        suffix.startswith(".") and hostname.endswith(suffix)
        for suffix in self.allowed_hosts
    )
```

The `.` prefix convention is well-established (it mirrors cookie domain scoping in RFC 6265 and DNS zone delegation) and is unambiguous: existing exact-match entries are unaffected, so this is fully backward-compatible.

**Describe alternatives you've considered**

1. **Glob/regex patterns** (e.g. `*.fabric.microsoft.com`): More expressive, but increases attack surface. A malformed regex or overly broad glob could inadvertently match unintended hosts. Suffix matching with the `.` prefix convention is simpler, harder to misconfigure, and covers the primary use case.

2. **Callback-based validation** (e.g. passing a `Callable[[str], bool]`): Maximum flexibility, but shifts security responsibility entirely to the consumer and makes misconfiguration easy. It would also break the current simple list-based API contract.

3. **Empty `allowed_hosts` (disable validation)**: Works today but is a security anti-pattern — bearer tokens would be sent to *any* host the request is redirected to or misconfigured against.

4. **Monkey-patching `is_url_host_valid`**: This is our current workaround. It works, but it's fragile, invisible to other kiota consumers, and could break with internal refactors.

**Additional context**

This is not limited to Microsoft Fabric. Any service that uses tenant-specific or resource-specific subdomains has the same problem. Examples include:

- **Microsoft Fabric GraphQL**: `<guid>.z2c.graphql.fabric.microsoft.com`
- **Azure API Management**: `<service-name>.azure-api.net`
- **Power BI Embedded**: `<cluster>.pbidedicated.windows.net`

The `AllowedHostsValidator` is a security boundary — it controls where bearer tokens are sent. Supporting suffix matching preserves that security role (tokens are still restricted to a known domain) while accommodating the reality of modern cloud APIs with dynamic hostnames.

dabla:
Support wildcard patterns in the allowlist so trusted host groups can be managed without listing every subdomain explicitly. Keep strict host normalization and validation to preserve safe trust decisions.

## Overview

This PR extends `AllowedHostsValidator` to support wildcard-style host matching via suffix entries (for example, `.fabric.microsoft.com`) while preserving exact-match behavior.

This is needed for APIs that use dynamic or tenant-specific subdomains, where enumerating every hostname is not practical. With this change, callers can keep host validation enabled instead of falling back to an empty allowed_hosts list.

## Related Issue

Fixes [#700](https://github.com/microsoft/kiota-python/issues/700)

### Demo

```python
validator = AllowedHostsValidator([".fabric.microsoft.com"])

validator.is_url_host_valid("https://2c662c47.z2c.graphql.fabric.microsoft.com")  # True
validator.is_url_host_valid("https://fabric.microsoft.com")                         # False
```

### Notes

Suffix entries are intentionally subdomain-only:
- `.fabric.microsoft.com` matches `foo.fabric.microsoft.com`
- `.fabric.microsoft.com` does **not** match `fabric.microsoft.com`

Exact host entries continue to work as before.

## Testing Instructions

* Check out this branch
* From repo root, run:
  * `pwsh -File .\kiota-python.ps1 test`
* Or run package-scoped tests:
  * `Set-Location .\packages\abstractions`
  * `poetry install`
  * `poetry run pytest .\tests\authentication\test_allowed_hosts_validator.py`
* Verify expected behavior:
  * wildcard/suffix entries match dynamic subdomains
  * bare domain is rejected for suffix entries
  * matching is case-insensitive
  * `set_allowed_hosts(...)` keeps suffix behavior

### DIFF
--- a/packages/abstractions/kiota_abstractions/authentication/allowed_hosts_validator.py
+++ b/packages/abstractions/kiota_abstractions/authentication/allowed_hosts_validator.py
@@ -67,4 +67,11 @@ class AllowedHostsValidator:
             return False
         if not o.hostname:
             return False
-        return o.hostname.lower() in self.allowed_hosts
+        hostname = o.hostname.lower()
+        # Existing exact match
+        if hostname in self.allowed_hosts:
+            return True
+        # Suffix match for entries starting with "."
+        return any(
+            suffix.startswith(".") and hostname.endswith(suffix) for suffix in self.allowed_hosts
+        )
--- a/packages/abstractions/tests/authentication/test_allowed_hosts_validator.py
+++ b/packages/abstractions/tests/authentication/test_allowed_hosts_validator.py
@@ -62,11 +62,28 @@ class TestIsUrlHostValid:
         validator = AllowedHostsValidator(["example.com"])
         assert validator.is_url_host_valid("https://sub.example.com/path") is False
 
+    def test_returns_true_for_subdomain_matching_allowed_suffix(self):
+        validator = AllowedHostsValidator([".fabric.microsoft.com"])
+        assert validator.is_url_host_valid(
+            "https://abc.123.graphql.fabric.microsoft.com/path"
+        ) is True
+
+    def test_returns_false_for_bare_domain_when_allowed_as_suffix(self):
+        validator = AllowedHostsValidator([".fabric.microsoft.com"])
+        assert validator.is_url_host_valid("https://fabric.microsoft.com/path") is False
+
+    def test_suffix_host_matching_is_case_insensitive(self):
+        validator = AllowedHostsValidator([".Fabric.Microsoft.COM"])
+        assert validator.is_url_host_valid("https://ABC.z2c.graphql.fabric.microsoft.com/path") is True
+
     def test_allows_multiple_valid_hosts(self):
-        validator = AllowedHostsValidator(["example.com", "api.example.com"])
+        validator = AllowedHostsValidator(["example.com", "api.example.com", ".fabric.microsoft.com"])
         assert validator.is_url_host_valid("https://example.com/path") is True
         assert validator.is_url_host_valid("https://api.example.com/path") is True
         assert validator.is_url_host_valid("https://other.com/path") is False
+        assert validator.is_url_host_valid(
+            "https://abc.123.graphql.fabric.microsoft.com/path"
+        ) is True
 
     def test_handles_url_with_port(self):
         validator = AllowedHostsValidator(["example.com"])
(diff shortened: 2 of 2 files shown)


baywet (dismissed):
Thanks for the contribution!

Would you be willing to make the same contribution in other languages to align the behaviour?

baywet (approved):

Yes, in general on any Microsoft owned repository, workflows need an approval before they can run for security reasons.

Would you be willing to reproduce the change across the other repositories?

Issue #133, opened 2023-09-11
[BUG] `APIError` __str__ method throws an error in python 3.10 tests.


musale:
Failing tests in Python 3.10 due to an error caused when the 'error' attribute is fetched. https://github.com/microsoft/kiota-http-python/actions/runs/6149817965/job/16686450244?pr=171.

PR #724, opened 2026-09-07
fix: lead APIError's string form with the error message

### LINKED ISSUE #723: `APIError.__str__` starts with a newline and ignores the generated `primary_message`, so every error is titled "ODataError:" with nothing after it
**Environment**

- Python Version: 3.12.13
- kiota-abstractions version: 1.10.1 and 1.12.0 (same code on main), with msgraph-sdk 1.55.0 and 1.62.0
- OS: macOS

**Describe the bug**

`APIError.__str__` (`packages/abstractions/kiota_abstractions/api_error.py`) returns a multi-line string that starts with a newline and indentation, and prints `message: {self.message}`, which is `None` for every generated error model because the generator never sets `message`; it emits a `primary_message` property instead, and `__str__` does not look at it.

Two consequences. `logging.error("...: %s", err)` prints a blank line, then an indented block. And anything that titles an exception by its first line, Sentry does exactly `value.splitlines()[0]` (https://github.com/getsentry/sentry/blob/master/src/sentry/eventtypes/error.py), shows `ODataError:` with nothing after it for every Graph error. In one Sentry project every throttling, not-found and permission error from Graph is grouped under that empty title, and the actual message (`Application is over its MailboxConcurrency limit.`) only appears in the expanded body.

The dotnet abstraction gets this right: the generated `ODataError` overrides `Message => Error?.Message ?? string.Empty` (https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/main/src/Microsoft.Graph/Generated/Models/ODataErrors/ODataError.cs), so `ex.Message` and `ex.ToString()` lead with the server's message.

**To Reproduce**

No network, just the generated model:

```python
import importlib.metadata as md
import logging

from msgraph.generated.models.o_data_errors.main_error import MainError
from msgraph.generated.models.o_data_errors.o_data_error import ODataError

err = ODataError(
    error=MainError(code="ApplicationThrottled", message="Application is over its MailboxConcurrency limit."),
    response_status_code=429,
)
print(f"kiota-abstractions {md.version('microsoft-kiota-abstractions')}, msgraph-sdk {md.version('msgraph-sdk')}")
print("primary_message:", repr(err.primary_message))
print("str(err):", repr(str(err)))
print("first line, what Sentry and most log viewers show as the title:", repr(str(err).splitlines()[0]))
logging.basicConfig(format="%(levelname)s %(message)s")
logging.getLogger().error("Graph call failed: %s", err)
```

Output, identical on 1.10.1 + msgraph-sdk 1.55.0 and 1.12.0 + msgraph-sdk 1.62.0:

```
ERROR Graph call failed: 
        APIError
        Code: 429
        message: None
        error: MainError(additional_data={}, code='ApplicationThrottled', details=None, inner_error=None, message='Application is over its MailboxConcurrency limit.', target=None)
        
primary_message: 'Application is over its MailboxConcurrency limit.'
str(err): "\n        APIError\n        Code: 429\n        message: None\n        error: MainError(...)\n        "
first line, what Sentry and most log viewers show as the title: ''
```

**Expected behavior**

A single line that leads with the message, something like `APIError: Application is over its MailboxConcurrency limit. (status 429, code ApplicationThrottled)`, taking the text from `primary_message` when the subclass defines it and falling back to `message`. Keeping `Code:` and the nested `error` on further lines is fine, as long as line one carries the message. Happy to send the PR; it is a few lines in `api_error.py` plus a test.

**Additional context**

#133 touched `__str__` for a 3.10 attribute-access crash, the format itself is unchanged since the abstractions package was consolidated in 2024.

HardMax71:
## Overview

`APIError.__str__` returned a multi-line string that started with a newline and printed `message: None` for every generated error model, because the generator never sets `message`; it emits a `primary_message` property that `__str__` ignored. Anything that titles an exception by its first line (Sentry does exactly `value.splitlines()[0]`, most log viewers do the same) showed `ODataError:` with nothing after it, and `logging.error("...: %s", err)` printed a blank line followed by an indented block.

`primary_message` is now a property on `APIError` that returns `message`; generated error models already override it with the message from the API's error payload. `__str__` reads it (falling back to the class name) and appends the status code in parentheses. There is no `getattr` lookup any more, and the nested `error:` payload line is gone with it; a generated model can render its payload in its own `__str__` if wanted. For the throttling example from the issue:

```
Application is over its MailboxConcurrency limit. (status 429)
```

This matches the dotnet abstraction, where the generated error overrides `Message` with `Error?.Message`; Python cannot override the `message` dataclass field with a property, so the override lives on `primary_message`.

## Related Issue

Fixes #723

### Notes

The `getattr(self, "error", None)` guard from #133 is removed; `__str__` no longer reaches into subclass fields.

## Testing Instructions

* `cd packages/abstractions && pytest tests/test_api_error.py`: a bare message, the status suffix, the class-name fallback, `primary_message` defaulting to `message`, a generated-style subclass overriding `primary_message`, and that no variant starts with a blank line.
* The whole abstractions suite: 139 passed. yapf, isort and mypy clean; pylint reports only the pre-existing `.pylintrc` option warnings.


packages/abstractions/kiota_abstractions/api_error.py
```diff
@@ -11,16 +11,18 @@ class APIError(Exception):
...
+        # error payload on ``error``; the base class knows neither, so both are optional here.
+        # The first line leads with the message so log lines and error titles show it.
         error = getattr(self, "error", None)
+        message = getattr(self, "primary_message", None) or self.message or type(self).__name__
+        details = []
+        if self.response_status_code is not None:
+            details.append(f"status {self.response_status_code}")
+        code = getattr(error, "code", None)
+        if code:
+            details.append(f"code {code}")
```
baywet: I'm a bit confused, where is this code property coming from?

baywet (commented):
Thanks for the contribution!

packages/abstractions/kiota_abstractions/api_error.py
HardMax71: It is the `code` field of an OData error body, see [OData JSON Format v4.01, Error Response](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ErrorResponse). On Graph that is `MainError.code`, for example `ApplicationThrottled`. Kiota itself has no idea about it, so it seems like it doesn't not belong in the base class.

Removed in https://github.com/microsoft/kiota-python/pull/724/commits/54848f9906b9c9e7c089859d7c6eef94abead09c.

packages/abstractions/kiota_abstractions/api_error.py
```diff
@@ -11,16 +11,14 @@ class APIError(Exception):
     response_headers: Optional[dict[str, str]] = None
 
     def __str__(self) -> str:
+        # Generated error models carry the server's message on ``primary_message`` and the
+        # error payload on ``error``; the base class knows neither, so both are optional here.
+        # The first line leads with the message so log lines and error titles show it.
         error = getattr(self, "error", None)
+        message = getattr(self, "primary_message", None) or self.message or type(self).__name__
```
baywet: I'm also confused about those two fields: they are not present in the base class? or defined here?

HardMax71: Not on the base class, both come from the generated error models. `error` is the field the generator makes for the API's error schema, on Graph that is [`ODataError.error`](https://github.com/microsoftgraph/msgraph-sdk-python/blob/main/msgraph/generated/models/o_data_errors/o_data_error.py#L20). The base class already read it with `getattr` before this PR (added in #137). `primary_message` is a property the generator adds to every error model whose schema marks a field with `x-ms-primary-error-message` ([PythonRefiner.cs#L153](https://github.com/microsoft/kiota/blob/main/src/Kiota.Builder/Refiners/PythonRefiner.cs#L153), [CommonLanguageRefiner.cs#L1577](https://github.com/microsoft/kiota/blob/main/src/Kiota.Builder/Refiners/CommonLanguageRefiner.cs#L1577)), on Graph [here](https://github.com/microsoftgraph/msgraph-sdk-python/blob/main/msgraph/generated/models/o_data_errors/o_data_error.py#L59).

So the base class only uses them when the subclass has them. If you'd like to, I can add `primary_message` to `APIError` as a property that returns `self.message`, the generated classes already override it and `__str__` can then read it directly.

baywet: Thank you for the additional information.
I think the python implementation has strayed from the other languages here. The behaviour is typically for the rendering method to be overloaded and generated, see [this example](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/16db3a1fb7ca71332ede033bafe24354d9603f32/src/Microsoft.Graph/Generated/Models/ODataErrors/ODataError.cs#L42) which is based on [this class](https://github.com/microsoft/kiota-dotnet/blob/4bba97ace0926246ad0c05497c71dbd07854a034/src/abstractions/ApiException.cs#L31) which exposes additional properties (like the status code and response headers), but doesn't rely on reflection to format the message.

Unless Python has some peculiar language constraints I can't remember, we should probably align things here.

HardMax71: Agreed, let's align. Python has one constraint here: `message` is a dataclass field on `APIError`, and a subclass can't override a dataclass field with a property. The generated `__init__` sets `self.message`, and a property without a setter fails right there. That is why the generator emits the override under another name, `primary_message`.

So the Python version of the dotnet shape would be: `primary_message` becomes a real property on `APIError` (returns `self.message`), the generated classes keep overriding it, and `__str__` reads `self.primary_message` plus the status code. No `getattr`. I would also drop the `error` payload lookup from #137, same kind of reflection; if the payload should stay in `str()`, the generator can emit `__str__` on the model. If that is what you mean, I will push it.

baywet: I think we're on the same page, thanks for exploring this issue with me.

HardMax71: Pushed in 4276f2b: `primary_message` is a property on `APIError` that returns `message`, generated models keep overriding it, `__str__` reads it plus the status code. Both `getattr` lookups are gone, and with them the `error:` payload line. Checked against the generated Graph `ODataError`: `Application is over its MailboxConcurrency limit. (status 429)`.

baywet (approved):

PR #726, opened 2026-09-10
fix: give each request configuration and headers inspection handler its own HeadersCollection

### LINKED ISSUE #507: RequestConfiguration instances share the same HeadersCollection instance by default
The default for `RequestConfiguration.headers` is an instance of `HeadersCollection`. This is created at class definition time, meaning that all instances of `RequestConfiguration` share the same instance of `HeadersCollection` by default. This leads to conflict when different instances of `RequestConfiguration` add/remove headers.

https://github.com/microsoft/kiota-python/blob/806c946d3f9f8709a3e8807b4dafa413a221de97/packages/abstractions/kiota_abstractions/base_request_configuration.py#L16-L22

Instead, a factory should be used so that a new instance is created:

```python
from dataclasses import dataclass, field
from typing import Generic, Optional, TypeVar

QueryParameters = TypeVar("QueryParameters")

@dataclass
class RequestConfiguration(Generic[QueryParameters]):
    headers: HeadersCollection = field(default_factory=HeadersCollection)
    options: Optional[list[RequestOption]] = None
    query_parameters: Optional[QueryParameters] = None
```

The same/similar issue also affects `HeadersInspectionHandlerOption` init param defaults.

https://github.com/microsoft/kiota-python/blob/806c946d3f9f8709a3e8807b4dafa413a221de97/packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py#L10-L20

HardMax71:
## Overview

`RequestConfiguration.headers` defaulted to one `HeadersCollection()` evaluated at class definition, so every configuration built without `headers=` shared it and a header added for one request rode along on every other request in the process. `field(default_factory=HeadersCollection)` gives each instance its own, as proposed in the issue.

The same shape sat two levels deeper on the http package. `HeadersInspectionHandlerOption.__init__` used `HeadersCollection()` as the default for both `request_headers` and `response_headers`; the `if request_headers else HeadersCollection()` guard never replaced them because `HeadersCollection` has no `__len__`, so an empty collection is truthy. And `HeadersInspectionHandler.__init__` used `HeadersInspectionHandlerOption()` as its default, so every handler built without options shared one option object and, through it, the same two collections: the headers one client inspected showed up on, and were cleared by, another. Both defaults are `None` now and the constructor creates a fresh object.

## Related Issue

Fixes #507

## Notes

The other middleware constructors (`RedirectHandler`, `RetryHandler`, `UserAgentHandler`, `UrlReplaceHandler`, `ParametersNameDecodingHandler`) take their option object as a function default too, so all handlers of one class share one option. Those options carry configuration values rather than per-request state, so the effect is limited to mutating `handler.options` after construction. Left out of this PR to keep it on #507; happy to send the same `Optional[...] = None` change for them separately.

## Testing Instructions

* `cd packages/abstractions && pytest tests/test_base_request_configuration.py`: two configurations get distinct collections, a header added to one is not visible on the other, a fresh configuration starts empty. Fails on main.
* `cd packages/http/httpx && pytest tests/middleware_tests/test_headers_inspection_handler.py`: two options do not share request or response collections, two handlers do not share an option object. Both fail on main.
* Full suites: abstractions 135 passed, http 110 passed. yapf, mypy and pylint clean on the touched source files.

### DIFF
--- a/packages/abstractions/tests/test_base_request_configuration.py
+++ b/packages/abstractions/tests/test_base_request_configuration.py
@@ -1,6 +1,9 @@
 import pytest
 
-from kiota_abstractions.base_request_configuration import BaseRequestConfiguration
+from kiota_abstractions.base_request_configuration import (
+    BaseRequestConfiguration,
+    RequestConfiguration,
+)
 
 
 def test_base_request_configuration_deprecation_warning():
@@ -11,3 +14,14 @@ def test_base_request_configuration_deprecation_warning():
 def test_import_base_request_configuration_no_warning():
     from kiota_abstractions.base_request_configuration import BaseRequestConfiguration, RequestConfiguration
     assert len(pytest.warns()) == 0
+
+
+def test_request_configurations_do_not_share_a_headers_collection():
+    first = RequestConfiguration()
+    second = RequestConfiguration()
+
+    first.headers.add("Prefer", "outlook.body-content-type=text")
+
+    assert first.headers is not second.headers
+    assert second.headers.try_get("Prefer") is False
+    assert RequestConfiguration().headers.count() == 0
--- a/packages/abstractions/tests/test_base_request_configuration.py
+++ b/packages/abstractions/tests/test_base_request_configuration.py
@@ -1,5 +1,9 @@
+import importlib
+import warnings
+
 import pytest
 
+from kiota_abstractions import base_request_configuration
 from kiota_abstractions.base_request_configuration import (
     BaseRequestConfiguration,
     RequestConfiguration,
@@ -12,8 +16,9 @@ def test_base_request_configuration_deprecation_warning():
 
 
 def test_import_base_request_configuration_no_warning():
-    from kiota_abstractions.base_request_configuration import BaseRequestConfiguration, RequestConfiguration
-    assert len(pytest.warns()) == 0
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        importlib.reload(base_request_configuration)
 
 
 def test_request_configurations_do_not_share_a_headers_collection():
--- a/packages/http/httpx/kiota_http/middleware/headers_inspection_handler.py
+++ b/packages/http/httpx/kiota_http/middleware/headers_inspection_handler.py
@@ -4,6 +4,8 @@
 # See License in the project root for license information.
 # ------------------------------------
 
+from typing import Optional
+
 from kiota_abstractions.request_option import RequestOption
 
 import httpx
@@ -21,16 +23,16 @@ class HeadersInspectionHandler(BaseMiddleware):
 
     def __init__(
         self,
-        options: HeadersInspectionHandlerOption = HeadersInspectionHandlerOption(),
+        options: Optional[HeadersInspectionHandlerOption] = None,
     ):
         """Create an instance of HeadersInspectionHandler
 
         Args:
             options (HeadersInspectionHandlerOption, optional): Default options to apply to the
-            handler. Defaults to HeadersInspectionHandlerOption().
+            handler. A new HeadersInspectionHandlerOption per handler when not provided.
         """
         super().__init__()
-        self.options = options
+        self.options = options if options is not None else HeadersInspectionHandlerOption()
 
     async def send(
         self, request: httpx.Request, transport: httpx.AsyncBaseTransport
--- a/packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
+++ b/packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
@@ -3,80 +3,34 @@
 # Licensed under the MIT License.
 # See License in the project root for license information.
 # ------------------------------------
-from typing import Optional
+from dataclasses import dataclass, field
+from typing import ClassVar
 
 from kiota_abstractions.headers_collection import HeadersCollection
 from kiota_abstractions.request_option import RequestOption
 
 
+@dataclass(eq=False)
 class HeadersInspectionHandlerOption(RequestOption):
-    """Config options for the HeaderInspectionHandler"""
+    """Config options for the HeaderInspectionHandler.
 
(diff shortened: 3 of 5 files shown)


packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
```diff
@@ -16,8 +18,8 @@ def __init__(
         self,
         inspect_request_headers: bool = True,
         inspect_response_headers: bool = True,
-        request_headers: HeadersCollection = HeadersCollection(),
-        response_headers: HeadersCollection = HeadersCollection(),
+        request_headers: Optional[HeadersCollection] = None,
```
baywet: since the default for the inspection is true, which is what most people are going to use, I suggest we initialize the collections from the beginning. What do you think?

packages/http/httpx/kiota_http/middleware/headers_inspection_handler.py
```diff
@@ -21,16 +23,16 @@ class HeadersInspectionHandler(BaseMiddleware):
 
     def __init__(
         self,
-        options: HeadersInspectionHandlerOption = HeadersInspectionHandlerOption(),
+        options: Optional[HeadersInspectionHandlerOption] = None,
```
baywet, line 26: same comment about initializing by default

baywet (changes requested):
Thanks for the contribution!

packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
HardMax71: They are. `__init__` creates a new `HeadersCollection` for each argument that was not passed, before anything else runs. The `= None` is only there because a Python default argument is evaluated once, when the function is defined. That is the shared instance from #507. The dataclass side of this PR does the same thing with `field(default_factory=HeadersCollection)`. I can write the body as `request_headers or HeadersCollection()` if you prefer it that way.

packages/http/httpx/kiota_http/middleware/headers_inspection_handler.py
HardMax71: Same here: the handler creates its own `HeadersInspectionHandlerOption` in `__init__` when none is passed. The `= None` avoids the one shared option object from #507.

packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
baywet: is there any downside to aligning on using field as a convention? besides maybe some additional memory pressure on the non canonical cases? At least the code would be consistent this way.

HardMax71: No downside I can find. `default_factory` runs only when the argument is left out, so passing your own collection costs nothing extra. Options are stored by `get_key()` and never compared or hashed, so `@dataclass(eq=False)` keeps the current behaviour, and `option.request_headers` stays the same public API.

Done in 4e54270: the option is a dataclass with `field(default_factory=HeadersCollection)`. Two notes: it is now the only dataclass among the nine option classes, I can convert the other eight and the five handlers with the same shared default in a follow-up. The handler itself keeps `= None`, it is a middleware with `super().__init__()`, not a dataclass.

packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
```diff
@@ -3,70 +3,34 @@
...
-        self._request_headers = value
-
-    @property
-    def response_headers(self):
-        """Gets the response headers to for the current request."""
-        return self._response_headers
-
-    @response_headers.setter
-    def response_headers(self, value: HeadersCollection):
-        self._response_headers = value
```
baywet, line 69: isn't removing the accessors going to be a source breaking change?

HardMax71: For plain use no: a dataclass field is an attribute, so reading and assigning `option.request_headers` works as before. It does break two things: a subclass that overrides one of these names as a property no longer constructs (the dataclass `__init__` assigns the attribute, and a property without a setter fails there), and the property objects are gone from the class. A dataclass cannot keep a property and a `field` under the same name, and renaming the fields would break keyword callers instead.

So I would go back to the plain class with the `None` default and keep the accessors, and leave `field(default_factory)` to `RequestConfiguration`, which is already a dataclass. Shall I revert 4e54270?

baywet (dismissed):

packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
baywet: I had missed the dataclass at the top. Thanks for confirming

packages/abstractions/tests/test_base_request_configuration.py
```diff
@@ -11,3 +14,14 @@ def test_base_request_configuration_deprecation_warning():
 def test_import_base_request_configuration_no_warning():
     from kiota_abstractions.base_request_configuration import BaseRequestConfiguration, RequestConfiguration
     assert len(pytest.warns()) == 0
```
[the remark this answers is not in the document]
HardMax71: Right, it asserted nothing. In 2789067 the test reloads the module under `warnings.simplefilter("error")`, so an import-time warning now fails it (checked by adding one).

packages/http/httpx/kiota_http/middleware/options/headers_inspection_handler_option.py
```diff
@@ -3,70 +3,34 @@
...
+    """Config options for the HeaderInspectionHandler.
+
```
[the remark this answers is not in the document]
HardMax71: Fixed in 2789067.

baywet (approved):

PR #730, opened 2026-09-11
fix(serialization-json): serialize nested arrays recursively

### LINKED ISSUE #478: [Bug] JSON Serialization Writer write_any_value does not handle nested lists
https://github.com/microsoft/kiota-python/blob/main/packages/serialization/json/kiota_serialization_json/json_serialization_writer.py#L473-L489

When sending the value [[1,2,3]] an error will be raised.

rksharma-owg:
`write_any_value("matrix", [[1, 2, 3]])` raises `TypeError` because the list dispatcher has no branch for nested lists.

Add an explicit `all(isinstance(x, list) for x in value)` branch, consistent with the neighboring dictionary check, and delegate to the existing recursive element writer. Keep the existing unknown-collection exception and specialized collection paths. This supports nested lists without accepting previously unsupported mixtures of lists and scalar/model/dictionary values at the same level.

Validation (Python 3.12):
- Regression coverage includes root/property values, additional data, empty arrays, dictionaries, model objects, and dates within nested lists.
- Three rejection cases failed before the review update and pass after it, confirming mixed collections retain the existing `TypeError`.
- JSON package: `pytest -q` — 120 passed.
- YAPF, isort, mypy, and `git diff --check` passed.
- Pylint rated the package 10/10 and reports the existing unrecognized `suggestion-mode` configuration option.

Closes #478.

### DIFF
--- a/packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
+++ b/packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
@@ -487,8 +487,13 @@ class JsonSerializationWriter(SerializationWriter):
                 self.write_collection_of_primitive_values(key, value)
             elif all(isinstance(x, dict) for x in value):
                 self.__write_collection_of_dict_values(key, value)
-            else:
+            elif all(isinstance(x, list) for x in value):
                 self.write_collection_of_primitive_values(key, value)
+            else:
+                raise TypeError(
+                    f"Encountered an unknown collection type during serialization {type(value)}\
+                    with key {key}"
+                )
         elif isinstance(value, dict):
             self.__write_dict_value(key, value)
         else:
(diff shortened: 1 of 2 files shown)


packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
```diff
@@ -488,10 +488,7 @@ def write_any_value(self, key: Optional[str], value: Any) -> Any:
             elif all(isinstance(x, dict) for x in value):
                 self.__write_collection_of_dict_values(key, value)
             else:
```
baywet: can we maintain the exception and test the instance like we're doing for dictionary above instead?

baywet (changes requested):
Thanks for the contribution!

packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
rksharma-owg: Thanks, that keeps the change better scoped. Added an explicit all(isinstance(x, list) for x in value) branch and restored the original collection TypeError in 326ebc5. Updated the regression cases to preserve mixed-collection rejection; all 120 JSON package tests pass, along with format, import and type checks.

baywet (approved):

Issue #365, opened 2022-12-01
Investigate whether the async credentials should be closed by this library or by the caller


baywet:
[Their documentation](https://learn.microsoft.com/en-us/python/api/azure-core/azure.core.credentials_async.asynctokencredential?view=azure-python) is terrible but I'm wondering what close does and whether we need to call it after each authentication request or when we don't need the credential object anymore? (wondering whether we'll have the same issue as the http package where it stops working after the first request)

_Originally posted by @baywet in https://github.com/microsoft/kiota-authentication-azure-python/pull/45#discussion_r1037099637_

PR #732, opened 2026-09-11
fix(authentication-azure): preserve caller-owned async credentials

### LINKED ISSUE #528: Async `AzureIdentityAccessTokenProvider` closes credential after every `get_token` breaking concurrent requests
Under concurrency (e.g. `asyncio.gather`), token acquisition fails with errors such as “HTTP transport has already been closed” / “Session closed”. Example:

```python
from azure.identity.aio import ClientSecretCredential
from msgraph import GraphServiceClient

import asyncio


credential = ClientSecretCredential(...)

client = GraphServiceClient(
    credentials=credential,
    scopes=["https://graph.microsoft.com/.default"],
)


async def fetch_users():
    await client.users.get()


async def main():
    await asyncio.gather(*[fetch_users() for i in range(5)])


if __name__ == "__main__":
    asyncio.run(main())
```

I guess the lifecycle of credentials object belongs to the app and we shouldn't implicitly close the transport. Suggested fix is to remove [this line](https://github.com/microsoft/kiota-python/blob/b7e3f4ce9c8d9354bf5832191ddada6e658d2d92/packages/authentication/azure/kiota_authentication_azure/azure_identity_access_token_provider.py#L118).

rksharma-owg:
`AzureIdentityAccessTokenProvider` closes an async credential immediately after retrieving a token. Reusing that credential for a later request, sharing it across providers, or acquiring tokens concurrently can then fail because its transport is already closed.

Leave the caller-owned credential open after token acquisition, matching the synchronous path. Document that the application must close it after all clients have finished, using its async context manager or `await credential.close()`.

Validation (Python 3.12):
- Three regressions failed on the original code: sequential requests, two providers sharing one credential, and concurrent requests. Tests use a credential double with an observable closed state and exercise the real provider; no live Azure credentials or network calls are needed.
- Authentication Azure package: `pytest -q` — 21 passed.
- YAPF, isort, mypy, wheel build, and `git diff --check` passed.
- Pylint exited successfully with 10/10; it also reports the existing unrecognized `suggestion-mode` configuration option.

Closes #528. Related lifetime discussion: #365.

### DIFF
--- a/packages/authentication/azure/README.md
+++ b/packages/authentication/azure/README.md
@@ -17,6 +17,12 @@ In order to use this library, install the package by running:
 pip install microsoft-kiota-authentication-azure
 ```
 
+The application owns the credential passed to `AzureIdentityAccessTokenProvider`.
+The provider leaves it open so it can be reused for token refreshes, concurrent
+requests, or multiple clients. Close the credential after all clients using it
+have finished. For async Azure Identity credentials, use an `async with` context
+manager around the client lifetime or call `await credential.close()` at shutdown.
+
 ## Contributing
 
 This project welcomes contributions and suggestions.  Most contributions require you to agree to a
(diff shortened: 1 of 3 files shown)


baywet (approved):
Thanks for the contribution!

PR #734, opened 2026-09-12
fix(http): close the wrapped async transport

### LINKED ISSUE #494: [Bug] AsyncKiotaTransport does not cleanup resources correctly
**Is your feature request related to a problem? Please describe.**
`AsyncKiotaTransport` might not be cleaning up resources correctly.

```
class AsyncKiotaTransport(httpx.AsyncBaseTransport):
    """A custom transport that implements Kiota middleware functionality
    """

    def __init__(self, transport: httpx.AsyncBaseTransport, pipeline: MiddlewarePipeline) -> None:
        self.transport = transport
        self.pipeline = pipeline

    async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
        if self.pipeline:
            response = await self.pipeline.send(request)
            return response

        response = await self.transport.handle_async_request(request)
        return response
```

`AsyncKiotaTransport` inherits methods like `__aexit__` and `aclose` from the `AsyncBaseTransport`. Then,   `KiotaClientFactory` uses `_load_middleware_to_client` and replaces the underlying `_transport` of the client with its `AsyncKiotaTransport`. Later calls to `httpx.AsyncClient.aclose()` will try and call `client._transport.aclose()` which will render it a no-op.

```
 def _load_middleware_to_client(
        client: httpx.AsyncClient, middleware: Optional[list[BaseMiddleware]]
    ) -> httpx.AsyncClient:
        current_transport = client._transport
        client._transport = KiotaClientFactory._replace_transport_with_custom_kiota_transport(
            current_transport, middleware
        )
        if client._mounts:
            mounts: dict = {}
            for pattern, transport in client._mounts.items():
                if transport is None:
                    mounts[pattern] = None
                else:
                    mounts[pattern
                           ] = KiotaClientFactory._replace_transport_with_custom_kiota_transport(
                               transport, middleware
                           )
            client._mounts = dict(sorted(mounts.items()))
        return client
```
**Describe the solution you'd like**
`AsyncKiotaTransport` should in turn properly cleanup the `self.transport` resource calling its `aclose()` method

**Describe alternatives you've considered**
NA
**Additional context**
We discovered the issue as part of an internal investigation regarding connection leaks on a Flask server hosting an agentic pipeline with Langgraph. We invoked a graph asynchronously with a kiota client (that was properly closed in code), but yet we are noticing connection leaks. We would like the opinion of the maintainers on the matter

rksharma-owg:
Closing a client created by `KiotaClientFactory` currently leaves its underlying transport open. `AsyncKiotaTransport` inherits the no-op `AsyncBaseTransport.aclose()`, so both explicit client closure and async context exit stop at the wrapper.

Delegate `aclose()` to the wrapped transport. This restores the HTTPX client lifecycle contract without changing request processing or credential ownership. Cleanup exceptions propagate to the caller.

Validation on Python 3.13.14:

- Five regression cases failed on upstream before the fix. They cover explicit closure, context exit, mounted transports, request exceptions, cancellation, repeated client closure, and cleanup errors using the real HTTPX client and Kiota factory.
- `pytest -q tests/test_kiota_client_factory.py`: 15 passed.
- `pytest -q`: 115 passed; upstream baseline was 110 passed.
- `yapf -dr kiota_http`, `mypy kiota_http`, and `git diff --check`: passed. Mypy checked 24 source files.
- `pylint kiota_http --disable=W --rcfile=.pylintrc`: exit 0, 10/10, with the existing unrecognized `suggestion-mode` option diagnostic.
- CI's `isort kiota_http`: exit 0. A supplementary `--check-only` run identifies one existing blank-line issue in untouched `redirect_handler_option.py`, also reproduced from the upstream file. No unrelated formatting change is included; the changed implementation passes the import check.
- `uv build --wheel --out-dir <external-artifacts-directory>`: passed.

Tests use in-memory transports and no live credentials or external HTTP service. Other Python versions and platforms were not run locally.

Closes #494.

### DIFF
--- a/packages/http/httpx/kiota_http/middleware/async_kiota_transport.py
+++ b/packages/http/httpx/kiota_http/middleware/async_kiota_transport.py
@@ -18,3 +18,7 @@ class AsyncKiotaTransport(httpx.AsyncBaseTransport):
 
         response = await self.transport.handle_async_request(request)
         return response
+
+    async def aclose(self) -> None:
+        """Close the wrapped transport."""
+        await self.transport.aclose()
--- a/packages/http/httpx/tests/test_kiota_client_factory.py
+++ b/packages/http/httpx/tests/test_kiota_client_factory.py
@@ -1,3 +1,5 @@
+import asyncio
+
 import httpx
 import pytest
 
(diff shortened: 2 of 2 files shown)


baywet (approved):
Thanks for the contribution!

PR #735, opened 2026-09-13
fix(http): end redirect and CAE retry spans

### LINKED ISSUE #508: httpx: OpenTelemetry spans never end for RedirectHandler_send and retry_cae_response_if_required
The OpenTelemetry spans created in `RedirectHandler.send` ("RedirectHandler_send - redirect N") and `HttpxRequestAdapter.retry_cae_response_if_required` are not ended on the normal (non-error) code paths. In trace viewers, these spans appear to run indefinitely, making requests look “hung” even when the underlying HTTP call completes.

To Reproduce: make any request using a kiota-generate client ?

Expected behavior: Spans for redirect attempts and CAE retry decision should end when the operation finishes, so durations reflect real time spent.

Suggested fix
- Ensure `_redirect_span.end()` is called for each redirect attempt, e.g. in a finally: around the loop body.
- Ensure the span started in `retry_cae_response_if_required` is ended on all return paths (including when it triggers a retry).

rksharma-owg:
## Overview

Redirect-attempt spans and CAE retry-decision spans remain open after successful HTTP operations because their normal exit paths never end them. End each span in a `finally` block so returns, redirects, transport errors, and cancellation all release it. Record the existing redirect-limit exception before ending its span, and keep the CAE span open until its awaited retry finishes.

Regression tests use real OpenTelemetry spans with an in-memory exporter. They cover successful and disabled redirects, redirect chains and limits, transport failures and cancellation, each no-retry CAE path, and successful or interrupted retries. The existing CAE integration test also verifies both recursive decision spans are exported.

## Related Issue

Fixes #508

## Testing Instructions

From `packages/http/httpx`, install `requirements-dev.txt`, then run:

- `pytest`: 124 tests passed on Python 3.12.
- `pytest tests/middleware_tests/test_redirect_handler.py tests/test_httpx_request_adapter.py -q`: 77 passed. All 14 added regression cases failed against the unmodified implementation.
- `yapf -dr kiota_http`, `pylint kiota_http --disable=W --rcfile=.pylintrc`, and `mypy kiota_http`: exited successfully; mypy checked 24 source files.
- `isort --check-only kiota_http/httpx_request_adapter.py kiota_http/middleware/redirect_handler.py`: passed.
- `uv build`: source distribution and wheel built successfully.

The package-wide `isort --check-only kiota_http` reports an existing import-order issue in unchanged `middleware/options/redirect_handler_option.py`; the same check fails on clean main at `1d104f2`. Pylint emits the same existing `suggestion-mode` configuration diagnostic on main and this branch while exiting zero. Neither baseline issue is changed here.


baywet (approved):
Thanks for the contribution!

rksharma-owg:
Cool..

PR #736, opened 2026-09-14
feat(http): implement body inspection handler

### LINKED ISSUE #418: http - implement body inspection handler
specification https://github.com/microsoftgraph/msgraph-sdk-design/pull/116
make sure you update the [documentation](https://learn.microsoft.com/en-us/openapi/kiota/middleware?tabs=csharp#middleware-handlers) as well

rksharma-owg:
Resolves #418

### Description
Implements the `BodyInspectionHandler` middleware and `BodyInspectionHandlerOption` for Kiota Python, following the shared middleware specification in microsoftgraph/msgraph-sdk-design#116.

### Summary of Changes
- Adds opt-in request and response body capture with byte and rewound `BytesIO` accessors.
- Registers `BodyInspectionHandler` in the default middleware pipeline.
- Preserves request and response payloads for downstream consumers, including replayable raw streamed responses.
- Carries request options through redirects using native HTTPX request extensions, while retaining the existing middleware compatibility bridge.
- Clears captured state before each use and isolates captures by execution context when concurrent requests share a handler-level option.
- Restores HTTPX download-byte accounting, decoded-content state, and decoder lifecycle so replayed compressed streams are consumed and counted normally.
- Leaves already-consumed responses without cached content and closed unreadable responses unchanged instead of raising during inspection.
- Avoids initializing or copying capture state on the default-disabled path while still clearing reused captures.
- Retains the original lifecycle state for already-buffered responses instead of exposing decoded bytes as a raw stream.
- Rejects a null request with an intentional `TypeError` before option or span access.
- Emits the standardized `com.microsoft.kiota.handler.bodyInspection.enable` telemetry attribute and keeps the span open through inspection, transport execution, response handling, and exception cleanup.
- Adds regression coverage for capture behavior, concurrent requests, byte accounting, request and response streaming, redirect propagation, state reset, null-request validation, span lifecycle, telemetry, and client factory integration.

### Documentation
- MicrosoftDocs/openapi-docs#175 updates the Microsoft Learn support table and is gated on this PR.

### Verification
- `pytest` in `packages/http/httpx`: 154 passed.
- Focused body-inspection tests: 25 passed.
- Redirect, adapter, and body-inspection tests: 102 passed.
- YAPF and targeted isort checks passed for changed files.
- Pylint rated the package 10.00/10; the repository emits its existing `suggestion-mode` configuration warning.
- Mypy reported no issues in 26 source files.
- Fork preflight passed the full Python 3.10-3.14 matrix for commit `eaa50db`: https://github.com/rksharma-owg/kiota-python/actions/runs/35131242553
- SonarCloud reports 0 open or confirmed new issues for commit `eaa50db`.


packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,90 @@
...
+from .middleware import BaseMiddleware
+from .options import BodyInspectionHandlerOption
+
+BODY_INSPECTION_KEY = "com.microsoft.kiota.handler.body_inspection.enable"
+
+
+class BodyInspectionHandler(BaseMiddleware):
+    """The Body Inspection Handler allows the developer to inspect the body of the
+    request and response.
+    """
```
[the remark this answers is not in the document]
baywet: @rksharma-owg if you want to update the documentation as well, the repository is located here [https://github.com/MicrosoftDocs/openapi-docs](https://github.com/MicrosoftDocs/openapi-docs/)

baywet:
@rksharma-owg would you mind addressing the copilot comments please?

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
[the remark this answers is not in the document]
rksharma-owg: Opened MicrosoftDocs/openapi-docs#175 to mark Python body inspection as part of the default middleware pipeline. The docs PR is draft and explicitly gated on this implementation PR.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,90 @@
...
+
+        Returns:
+            BodyInspectionHandlerOption: The options to be used.
+        """
+        current_options = None
+        request_options = getattr(request, "options", None)
+        if request_options:
+            current_options = request_options.get(BodyInspectionHandlerOption.get_key(), None)
+        if current_options:
+            return current_options
```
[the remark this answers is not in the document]
rksharma-owg: Redirect-created requests now receive a copy of the original request options in ee0ea54. The regression test verifies that the same per-request body inspection option captures the final redirected response.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,90 @@
...
+# ------------------------------------
+
+from typing import Optional
+
+import httpx
+
+from .middleware import BaseMiddleware
+from .options import BodyInspectionHandlerOption
+
+BODY_INSPECTION_KEY = "com.microsoft.kiota.handler.body_inspection.enable"
```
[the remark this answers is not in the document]
rksharma-owg: Updated the attribute to `com.microsoft.kiota.handler.bodyInspection.enable` in ee0ea54 and added a regression test for the emitted key.

rksharma-owg:
Addressed the review findings in ee0ea54: standardized the telemetry key, preserved per-request options across redirects, restored replayable raw response streams after inspection, and added regressions for each behavior. The full Python 3.10-3.14 fork preflight passed, and the linked documentation update is MicrosoftDocs/openapi-docs#175.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        """To execute the current middleware
+
+        Args:
+            request (httpx.Request): The prepared request object
+            transport (httpx.AsyncBaseTransport): The HTTP transport to use
+
+        Returns:
+            httpx.Response: The response object.
+        """
+        current_options = self._get_current_options(request)
```
baywet, line 50: The .NET implementation explicitly rejects a null request before reading options: https://github.com/microsoft/kiota-dotnet/blob/e16c78c20b4c2bb3399888d73b68ae735aa6320c/src/http/httpClient/Middleware/BodyInspectionHandler.cs#L39-L40. Should we mirror that here with an explicit guard (for example, raising ValueError/TypeError) so the failure mode is intentional instead of an attribute-access error later?

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        span = self._create_observability_span(request, "BodyInspectionHandler_send")
+        span.set_attribute(BODY_INSPECTION_KEY, True)
+        span.end()
```
baywet: The .NET implementation keeps the observability activity open for the full handler execution and disposes it in inally, covering request inspection, downstream send, response inspection, and exceptions: https://github.com/microsoft/kiota-dotnet/blob/e16c78c20b4c2bb3399888d73b68ae735aa6320c/src/http/httpClient/Middleware/BodyInspectionHandler.cs#L44-L78. Here the span is ended before any of that work runs, so it only records enablement. Should this span stay open around the send path instead?

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if current_options and current_options.inspect_request_body:
+            content = await request.aread()
+            if content:
```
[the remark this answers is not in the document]
rksharma-owg: Verified against every supported HTTPX minor release (0.25.0 through 0.28.1): `Request.aread()` replaces a non-`ByteStream` request stream with `ByteStream(self._content)`. I strengthened the regression in 1dc4c07 so the transport iterates `request.stream` directly instead of calling `request.read()`; the async-generator payload is preserved without duplicating HTTPX behavior in this middleware.

rksharma-owg:
Addressed the new review in 1dc4c07. Reused per-request options now clear stale captures, and already-buffered responses retain their original consumed/closed raw-stream state. The request-stream concern was verified against HTTPX 0.25.0-0.28.1 and is already handled by `Request.aread()`; the regression now exercises `request.stream` directly at the transport boundary. Local validation reports 133 tests passing, and the full Python 3.10-3.14 fork preflight passed: https://github.com/rksharma-owg/kiota-python/actions/runs/35100712297

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
rksharma-owg: Absolutely — thanks, Vincent. I agree the explicit failure is much clearer here. I added `TypeError("request cannot be null")` before option or span access, along with a focused regression test in 867f2e6.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
rksharma-owg: Good catch, Vincent — the .NET lifecycle is the right model here. I moved `span.end()` into a `finally`, so the span now covers request inspection, the downstream send, response inspection, and still closes if the transport raises. I added tests for both the normal path and exception cleanup in 867f2e6; the full Python 3.10–3.14 fork preflight is green.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,108 @@
...
+        current_options = None
+        request_options = getattr(request, "options", None)
+        if request_options:
+            current_options = request_options.get(BodyInspectionHandlerOption.get_key(), None)
+        if not current_options:
+            current_options = self.options
+
+        # Clear body per request
+        current_options.request_body = None
+        current_options.response_body = None
```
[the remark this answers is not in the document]
rksharma-owg: Good catch — addressed in `b80f4fc`. Captures on the original option are now isolated by execution context, so overlapping sends can safely share a handler-level option without clearing or overwriting each other's request/response bodies. I also added a forced-overlap regression test that verifies each task sees its own matching pair.

rksharma-owg:
Addressed both findings from the latest Copilot review in `b80f4fc`:

- handler-level capture state is now isolated per execution context, with a concurrent forced-overlap regression test;
- response restoration now also restores HTTPX byte accounting, with coverage confirming the count stays at its pre-inspection value until the caller consumes the replayed stream.

Local validation is green (`136` HTTP package tests; `21` focused tests), and the full Python 3.10–3.14 fork preflight passed: https://github.com/rksharma-owg/kiota-python/actions/runs/35104897911

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,112 @@
...
+                    current_options.request_body = content
+                else:
+                    current_options.request_body = None
+
+            response = await super().send(request, transport)
+
+            if current_options and current_options.inspect_response_body:
+                if response.is_stream_consumed:
+                    content = await response.aread()
+                    raw_content = content
```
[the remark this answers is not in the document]
rksharma-owg: Thanks for catching that distinction. Addressed in `ae25c49`: the handler now checks for cached content independently from `is_stream_consumed`. If a response was already consumed without a cache, inspection leaves it unchanged and records no response body instead of calling `aread()` and raising `StreamConsumed`. The new regression test exercises a fully consumed uncached response.

rksharma-owg:
I also addressed the decoded-stream lifecycle finding in `ae25c49`. Rewinding now clears the temporary decoded-content cache and HTTPX's stateful decoder before restoring the raw stream, so a caller using `aiter_bytes()` consumes the replayed stream normally, closes the response, and updates `num_bytes_downloaded` correctly. The regression uses a gzip stream to verify decoded content, lifecycle flags, and raw-byte accounting.\n\nValidation is green: `138` HTTP package tests, `23` focused body-inspection tests, HTTPX 0.25.0–0.28.1 compatibility checks, and the full Python 3.10–3.14 fork preflight: https://github.com/rksharma-owg/kiota-python/actions/runs/35109454408

rksharma-owg:
Thanks for the extra pass — both moderate findings are addressed in `22d08e5`:\n\n- capture state is now cleared only when a prior request/response body actually exists, so the default-disabled handler leaves the context map uninitialized while reused options still reset correctly;\n- response inspection now skips closed, uncached streams just as it skips consumed, uncached streams, preserving the original response instead of raising `StreamClosed`.\n\nI added focused regressions for both paths. Validation is green: `140` HTTP package tests, `25` focused body-inspection tests, and the full Python 3.10–3.14 fork preflight: https://github.com/rksharma-owg/kiota-python/actions/runs/35113744365

packages/http/httpx/kiota_http/middleware/redirect_handler.py
```diff
@@ -64,6 +64,7 @@ async def send(
         """
         _enable_span = self._create_observability_span(request, "RedirectHandler_send")
         current_options = self._get_current_options(request)
+        request_options = getattr(request, "options", None)
```
baywet: we shouldn't have to use reflection here?

rksharma-owg: You are right, Vincent — thanks for the nudge. I moved option propagation to HTTPX native request extensions in `eaa50db`, so the reflective read in `RedirectHandler` is gone. The existing `options` attribute remains only as a compatibility bridge for the current middleware, and the redirect regression now starts from the native extension to verify the option reaches the follow-up request. The full Python 3.10–3.14 fork preflight is green: https://github.com/rksharma-owg/kiota-python/actions/runs/35131242553

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
```diff
@@ -0,0 +1,116 @@
...
+        request options are passed.
+
+        Args:
+            request (httpx.Request): The prepared request object
+
+        Returns:
+            BodyInspectionHandlerOption: The options to be used.
+        """
+        current_options = None
+        request_options = getattr(request, "options", None)
```
baywet: same here, I don't feel we should need reflection

packages/http/httpx/kiota_http/middleware/redirect_handler.py
```diff
@@ -129,13 +136,18 @@ def _build_redirect_request(
         url = self._redirect_url(request, response, options)
         stream = self._redirect_stream(request, method)
 
+        new_request_options = request_options.copy() if request_options else {}
+        extensions = request.extensions.copy()
+        if request_options is not None:
+            extensions[REQUEST_OPTIONS_KEY] = new_request_options
```
baywet, line 142: why is the redirect handler using the same key for options as the body inspection one? wouldn't that lead to a confusion?

packages/http/httpx/kiota_http/middleware/redirect_handler.py
```diff
@@ -64,6 +64,7 @@ async def send(
         """
         _enable_span = self._create_observability_span(request, "RedirectHandler_send")
         current_options = self._get_current_options(request)
+        request_options = request.extensions.get(REQUEST_OPTIONS_KEY)
```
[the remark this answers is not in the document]
rksharma-owg: Addressed in 925196e: `RedirectHandler.send()` and `_get_current_options()` now check `request.extensions.get(REQUEST_OPTIONS_KEY)` first and fall back to `getattr(request, 'options', None)` when absent. The redirect builder also propagates the copied options to both the extension dictionary and `new_request.options`, and I added `test_legacy_request_options_apply_to_redirected_response` to ensure legacy attribute options remain effective across redirects.

packages/http/httpx/tests/middleware_tests/test_body_inspection_handler.py
```diff
@@ -0,0 +1,632 @@
...
+async def test_inspected_buffered_response_keeps_consumed_raw_stream_state():
+    """Ensures decoded buffered content is not exposed as replayable raw bytes."""
+    compressed_content = gzip.compress(b"decoded response")
+
+    def request_handler(request: httpx.Request):
+        return httpx.Response(
+            200,
+            headers={"Content-Encoding": "gzip"},
+            content=compressed_content,
+        )
```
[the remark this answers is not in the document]
rksharma-owg: Addressed in 925196e: updated the test mock handler to construct a streamed response (`stream=httpx.ByteStream(compressed_content)`) and call `response.read()`, accurately simulating an HTTPX response that has streamed and decoded the compressed payload while consuming its raw stream.

packages/http/httpx/kiota_http/middleware/body_inspection_handler.py
rksharma-owg: Addressed in 925196e: removed the reflective access from the primary path. `BodyInspectionHandler._get_current_options()` now reads `request.extensions.get(REQUEST_OPTIONS_KEY)` directly, falling back to `getattr(request, 'options', None)` only for backwards compatibility with legacy callers.

packages/http/httpx/kiota_http/middleware/redirect_handler.py
rksharma-owg: To clarify: `REQUEST_OPTIONS_KEY` (`"kiota_request_options"`) is the top-level key under HTTPX's native `request.extensions` dictionary that holds the map of all middleware request options (`dict[str, RequestOption]`), replacing the monkey-patched `request.options` attribute across the pipeline. Inside that dictionary, each middleware option still uses its own distinct key (such as `RedirectHandlerOption.get_key()` and `BodyInspectionHandlerOption.get_key()`), so individual handlers do not collide.

packages/http/httpx/kiota_http/middleware/redirect_handler.py
```diff
@@ -64,6 +64,9 @@ async def send(
...
+        if request_options is None:
+            request_options = getattr(request, "options", None)
```
baywet: we're still relying on reflection here, and a number of other places. Can you please clean it up?

rksharma-owg: Cleaned up reflection across all middleware handlers (`redirect_handler`, `body_inspection_handler`, `headers_inspection_handler`, `parameters_name_decoding_handler`, `retry_handler`, `url_replace_handler`, `user_agent_handler`, and `BaseMiddleware`). All handlers now read options directly from HTTPX's native extension dictionary via `request.extensions.get(REQUEST_OPTIONS_KEY)`, and monkey-patched `request.options` has been eliminated. All unit and middleware tests pass.

baywet:
@rksharma-owg can you run the formatting command please?

rksharma-owg:
okay

rksharma-owg:
Done, I ran the repository formatting command (`yapf -ir kiota_http`) and pushed the resulting formatting changes. I also reran the validation suite and confirmed everything is green:

- **YAPF check**: `poetry run yapf -dr kiota_http` passed cleanly (0 differences)
- **Pylint**: 10.00/10 rating
- **Mypy**: 0 issues across 26 source files
- **pytest**: 154 passed (including all 25 `BodyInspectionHandler` tests)
- **CI Preflight**: Full Python 3.10–3.14 matrix passed on fork run [35360699982](https://github.com/rksharma-owg/kiota-python/actions/runs/35360699982)

baywet (approved):

PR #740, opened 2026-09-17
Add support for the HTTP QUERY verb (RFC 10008)

ricardoboss:
## Overview

This adds support for the relatively new HTTP verb `QUERY`, which allows sending a request body (like `POST) while being able to be cached (unlike `POST`).

## Related Issue

Relates to https://github.com/microsoft/kiota/issues/8126


baywet (approved):

baywet (dismissed):

baywet:
Actually @ricardoboss re-reading the bot's comments, and the RFC, I think it's correct for the redirect handler, would you mind looking into this please? And we'll probably need to replicate the change across languages.

baywet (approved):

PR #743, opened 2026-09-21
fix(json): preserve falsy root values in serialized content

### LINKED ISSUE #742: JSON serializer replaces falsy root values with an empty object
### Describe the bug

On current main (aaad20dfa6fa815079ae6c06ffd351c06b75639c), writing a root JSON value of false, zero, an empty string or an empty list produces `{}` instead of that value. This also affects `RequestInformation.set_content_from_scalar`, so a generated client can send a different JSON type and value than the caller supplied.

### Reproduction

```python
from kiota_serialization_json.json_serialization_writer import JsonSerializationWriter

for value in [False, 0, 0.0, "", []]:
    writer = JsonSerializationWriter()
    writer.write_any_value(None, value)
    print(repr(value), writer.get_serialized_content())
```

All five outputs are `b'{}'`. Expected outputs are `b'false'`, `b'0'`, `b'0.0'`, `b'""'` and `b'[]'`, respectively.

`get_serialized_content()` checks the truthiness of `self.value` instead of whether a root value has been set. The same check also lets a falsy root value bypass the existing rejection of mixed root values and named properties.

Tested with Python 3.12 on Windows using the local JSON and abstractions packages. A two-condition fix and regression tests are prepared. This is separate from #741: that concerns empty *property names*, whereas this occurs with `key=None` and affects root *values*. Investigation and tests are AI-assisted.

Shubham-Padkonde:
## Overview

Root JSON values `false`, `0`, `0.0`, `""` and `[]` currently serialize as `{}`, including through `RequestInformation.set_content_from_scalar`. Explicit root `null` also needs to be distinguished from an unset writer. Track root presence separately from the payload, preserve it for composed objects, use it for mixed-root/property validation, and reset it after serialization.

## Related Issue

Fixes #742

## Testing Instructions

From `packages/serialization/json`, install `requirements-dev.txt` and run `pytest`.

- 148 package tests pass on Windows/Python 3.12.
- 28 added cases cover root round trips, writer reuse, mixed-root/property rejection, actual RequestInformation content, and composed-object null serialization. The original falsy-root regressions produced 16 failures before the initial fix; both added root-null cases failed before the review follow-up.
- Mypy (6 source files), YAPF and isort checks pass. The repository's pylint command exits successfully with 10.00/10, while reporting the existing unrecognized `suggestion-mode` configuration option.

## Notes

AI-assisted implementation and testing. The full multi-package and Python-version CI matrix was not run locally. This starts from main and is independent of #741, which fixes empty property names.


packages/serialization/json/tests/unit/test_root_values.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.mark.parametrize("value", [False, 0, 0.0, "", [], {}, True, 1, "text", [1]])
+def test_root_value_round_trip_and_reset(value):
+    writer = JsonSerializationWriter()
+    writer.write_any_value(None, value)
+    result = json.loads(writer.get_serialized_content())
+    assert type(result) is type(value)
+    assert result == value
+    writer.write_str_value("name", "next")
+    assert json.loads(writer.get_serialized_content()) == {"name": "next"}
```
baywet, line 22: what's the value of those last two lines?

Shubham-Padkonde:
Addressed the root-null finding by tracking root presence separately from its payload, including composed-object serialization, and clearing that state after serialization. The new root-null cases failed on the previous implementation; all 148 JSON tests now pass, along with mypy, YAPF, isort and the repository's pylint command (exit 0; existing suggestion-mode configuration diagnostic remains).

The last two lines in the round-trip test verify writer reuse: after serializing a root value, writing a named property must succeed without stale root state causing an invalid mixed-root error. I added a comment explaining that purpose. Agent-assisted update.

packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
```diff
@@ -24,13 +24,24 @@ class JsonSerializationWriter(SerializationWriter):
...
 
+    @property
+    def value(self) -> Any:
+        """The root payload, which may itself be JSON null."""
+        return self._value
+
+    @value.setter
+    def value(self, value: Any) -> None:
+        self._value = value
+        self._has_root_value = True
```
baywet: Here we're expending the public API surface and we shouldn't do that.

baywet (changes requested):

packages/serialization/json/kiota_serialization_json/json_serialization_writer.py
Shubham-Padkonde: Removed the property/setter and restored the existing value attribute. Root-writing methods now use a private _write_root_value helper to track presence; serialization resets the attribute and the private flag. All 148 JSON tests, mypy, formatting/import checks and the repository pylint command pass. Agent-assisted update.

baywet (approved):

